PR #2, opened 2025-11-26
Add ip rules configuration script

tsorya:
changing default to 4.20.4 
Adding systemd service that will wait till interface with local link that matches ovnk one will appear and configure ip rules

New OVNK/NetworkManager Issue:
OVNK is using a VF in order to create the ovn-k8s-int interface.
This interface is critical for OVNK functionality.
SRIOV is on charge of VFs and reconfigures them at some point
We have a configuration provided by machine-config-operator that sets unmanaged to this interface but it seems not to work.
NetworkManager sees that ovn-k8s-int interface is unmanaged and removes it.
We found a workaround by adding a harder NetworkManager configuration to unmanage this interface.
Another option is to run sriov as a systemd service, currently easier to add a new unmanaged configuration option.

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift release references and environment defaults to 4.20.4.
  * Adjusted a host-mode label value used for DPU detection.
  * Removed a prior declarative MachineConfig that disabled/masked certain network services.

* **New Features**
  * Added a pod-to-worker routing configuration script and a dedicated routing service for workers.
  * Enhanced worker bridge setup to include an unmanaged interface config and routing wiring.


romfreiman:
@coderabbitai review

killianmuldoon:
Looks good to me!

szigmon (commented):
Overall the approach is sound - using NetworkManager unmanaged config and policy routing to work around the OVNK/NM conflict. I have a few suggestions on specific lines below.

configuration_templates/configure-pod-to-worker-routing.sh
```diff
@@ -0,0 +1,101 @@
...
+
+CHECK_INTERVAL=2  # Check every 2 seconds
+PRIMARY_IP_FILE="/run/nodeip-configuration/primary-ip"
+
+echo "Waiting for primary IP file to determine IP version..."
+
+# Wait for primary-ip file and determine IP version first
+while [ ! -f "$PRIMARY_IP_FILE" ] || [ ! -s "$PRIMARY_IP_FILE" ]; do
+    sleep $CHECK_INTERVAL
+done
```
szigmon, line 13: test

szigmon (commented):
Review with inline suggestions

configuration_templates/configure-pod-to-worker-routing.sh
szigmon, line 13: **Suggestion: Add timeout to prevent infinite wait**

The script will wait forever if `/run/nodeip-configuration/primary-ip` never appears. Consider adding a timeout:

```bash
MAX_WAIT=300  # 5 minutes
ELAPSED=0
while [ ! -f "$PRIMARY_IP_FILE" ] || [ ! -s "$PRIMARY_IP_FILE" ]; do
    sleep $CHECK_INTERVAL
    ELAPSED=$((ELAPSED + CHECK_INTERVAL))
    if [ $ELAPSED -ge $MAX_WAIT ]; then
        echo "Error: Timed out waiting for primary IP file"
        exit 1
    fi
done
```

configuration_templates/configure-pod-to-worker-routing.sh
```diff
@@ -0,0 +1,101 @@
...
+            
+            # Get metric from br-dpu (default to 425 if not found)
+            br_dpu_metric=$(ip $IP_FLAG -j route show dev br-dpu | jq -r '.[] | select(.protocol == "kernel") | .metric // 425' | head -n1)
+            
+            echo ""
+            echo "Creating routing rules in table 100..."
+            
+            # 1. Add policy routing rule
+            echo "Adding rule: from $br_dpu_ip/$PREFIX_LEN lookup 100"
+            ip $IP_FLAG rule add from $br_dpu_ip/$PREFIX_LEN lookup 100 2>/dev/null || echo "Rule already exists or failed"
```
szigmon: **Suggestion: Distinguish 'already exists' from actual errors**

This suppresses both expected "already exists" errors AND unexpected failures (permissions, syntax issues). Consider checking if the rule exists first:

```bash
if ! ip $IP_FLAG rule show | grep -q "from $br_dpu_ip/$PREFIX_LEN lookup 100"; then
    ip $IP_FLAG rule add from $br_dpu_ip/$PREFIX_LEN lookup 100
fi
```

Same applies to the route commands on lines 88 and 92.

configuration_templates/configure-pod-to-worker-routing.sh
```diff
@@ -0,0 +1,101 @@
...
+            echo "Adding route: $ovnk_subnet via $br_dpu_gateway table 100"
+            ip $IP_FLAG route add $ovnk_subnet via $br_dpu_gateway table 100 2>/dev/null || echo "Route already exists or failed"
+            
+            # 3. Add br-dpu subnet route
+            echo "Adding route: $br_dpu_network dev br-dpu proto kernel scope link src $br_dpu_ip metric $br_dpu_metric table 100"
+            ip $IP_FLAG route add $br_dpu_network dev br-dpu proto kernel scope link src $br_dpu_ip metric $br_dpu_metric table 100 2>/dev/null || echo "Route already exists or failed"
+            
+            echo ""
+            echo "Routing configuration completed!"
+            exit 0
```
szigmon, line 111: **Question: Single interface assumption**

The script exits after processing the first OVN-K interface found. Is this intentional? If multiple OVN-K interfaces could exist in some topologies, this might need adjustment or at least documentation.

manifests/cluster-installation/99-worker-bridge.yaml
```diff
@@ -32,3 +42,27 @@ spec:
...
+        - name: ovs-configuration.service
+          enabled: false
+        - name: openvswitch.service
+          enabled: false
+          mask: true
+        - name: wait-for-br-ex-up.service
+          enabled: false
+        - name: p0-routing.service
+          enabled: true
+          contents: |
```
szigmon, line 54: **Suggestion: Add dependency on nodeip-configuration.service**

The script depends on `/run/nodeip-configuration/primary-ip` which is created by `nodeip-configuration.service`. Consider adding it to the dependencies:

```yaml
After=kubelet.service network-online.target nodeip-configuration.service
Wants=network-online.target nodeip-configuration.service
```

This would make the startup ordering more explicit and reliable.

manifests/cluster-installation/cno-dpu-host-cm.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
     name: hardware-offload-config
     namespace: openshift-network-operator
 data:
-    dpu-host-mode-label: "feature.node.kubernetes.io/dpu-enabled"
+    dpu-host-mode-label: "feature.node.kubernetes.io/dpu-enabled="
```
szigmon, line 7: **Verification needed: Trailing equals sign**

This changes the label format from `feature.node.kubernetes.io/dpu-enabled` to `feature.node.kubernetes.io/dpu-enabled=` (with trailing `=`). 

Please confirm this matches the expected CNO format - is this a label key with empty value, or was the `=` previously missing?

szigmon (commented):
Good approach to work around the OVNK/NetworkManager conflict. I have a few suggestions below.

configuration_templates/configure-pod-to-worker-routing.sh
tsorya: It is on purpose

tsorya: Test?

configuration_templates/configure-pod-to-worker-routing.sh
tsorya: Yes

manifests/cluster-installation/99-worker-bridge.yaml
tsorya: No need, kubelet depends on it

manifests/cluster-installation/cno-dpu-host-cm.yaml
tsorya: Wrong assumption, no need for trailing space

wabouhamad (approved):
/lgtm

PR #5, opened 2025-11-28
Fix broken ovnk-template + 4.19 workaround

tsorya:
## Summary by CodeRabbit

* **Chores**
  * Refactored OVN Kubernetes utilities image configuration to use a single placeholder for optional utils images.
  * Updated defaults so utils image repo/tag are optional (won’t be injected unless provided).
  * Improved manifest substitution to perform multiple replacements in one pass with clearer logging.
  * Made certain post-install manifests conditionally included based on environment flags and adjusted post-install flow timing.


scripts/utils.sh
```diff
@@ -332,15 +333,26 @@ update_file_multi_replace() {
 
     log [INFO] "Updating ${source_file} with multiple replacements..."
         
-    cp "${source_file}" "${target_file}"
+    # Read the entire source file content
+    local content
+    content=$(<"${source_file}")
```
szigmon, line 338: **Suggestion: Add source file validation**

The function should validate that the source file exists before attempting to read it. While `set -e` may catch this, explicit validation provides clearer error messages:

```bash
if [ ! -f "${source_file}" ]; then
    log "ERROR" "Source file not found: ${source_file}"
    return 1
fi
```

scripts/utils.sh
```diff
@@ -332,15 +333,26 @@ update_file_multi_replace() {
...
+    
     local i=0
     while [ $i -lt ${#pairs[@]} ]; do
         local placeholder="${pairs[$i]}"
         local value="${pairs[$((i+1))]}"
-        sed -i "s|${placeholder}|${value}|g" "${target_file}"
-        log [INFO] "Replaced ${placeholder} with ${value} in ${target_file}"
+        log [INFO] "Replacing ${placeholder} with ${value} in ${target_file}"
+
```
szigmon, line 345: **Minor: Log format inconsistency**

This uses `log [INFO]` but other functions in this file use `log "INFO"` (with quotes). Consider standardizing:

```bash
log "INFO" "Replacing ${placeholder} in ${target_file}"
```

Same applies to lines 333, 342, 350, and 352.

scripts/utils.sh
```diff
@@ -332,15 +333,26 @@ update_file_multi_replace() {
...
+        # Use bash parameter expansion for replacement (handles multi-line naturally)
+        content="${content//${placeholder}/${value}}"
+        
+        log [INFO] "Replaced ${placeholder} in ${target_file}"
         i=$((i+2))
     done
+    
+    # Write the modified content to target file
+    printf '%s\n' "${content}" > "${target_file}"
```
szigmon, line 354: **Good improvement!**

Using bash parameter expansion `${content//${placeholder}/${value}}` instead of `sed` correctly handles multi-line values. This is the right fix for YAML blocks like the `imagedpf:` section.

scripts/post-install.sh
```diff
@@ -80,6 +79,19 @@ function update_hbn_ovn_manifests() {
     
     # Update ovn-template.yaml for DPUDeployment
     if [ -f "${POST_INSTALL_DIR}/ovn-template.yaml" ]; then
+        # Determine the replacement value for <OVN_KUBERNETES_UTILS_IMAGES>
+        local utils_images_replacement=""
+        if [ -n "${OVN_KUBERNETES_UTILS_IMAGE_REPO}" ] && [ -n "${OVN_KUBERNETES_UTILS_IMAGE_TAG}" ]; then
+            # Build the imagedpf block with proper indentation (8 spaces for imagedpf:, 10 spaces for repository/tag)
```
szigmon, line 85: **Good: Dynamic imagedpf block construction**

Nice approach to conditionally build the `imagedpf:` block only when both repo and tag are set. The indentation (8 spaces for `imagedpf:`, 10 for children) matches the YAML structure correctly.

scripts/manifests.sh
```diff
@@ -133,9 +133,11 @@ function prepare_cluster_manifests() {
         "nfd-subscription.yaml"
         "sriov-subscription.yaml"
         "99-worker-bridge.yaml"
-        "4.19-cataloguesource.yaml"
     )
```
szigmon, line 136: **Question: Logic seems inverted?**

The condition adds `4.19-cataloguesource.yaml` to exclusions when `USE_V419_WORKAROUND` is **not** true. This means:
- `USE_V419_WORKAROUND=true` → file is **included**
- `USE_V419_WORKAROUND=false` → file is **excluded**

Is this the intended behavior? The naming suggests the workaround should be opt-in (excluded by default).

scripts/post-install.sh
```diff
@@ -235,7 +246,6 @@ function prepare_post_installation() {
         log [ERROR] "Post-installation directory not found: ${POST_INSTALL_DIR}"
```
szigmon, line 246: **Question: Why remove get_kubeconfig call?**

The `get_kubeconfig` call was removed from `prepare_post_installation`. Is this intentional? If `prepare_post_installation` can be called independently, the kubeconfig may not be available.

If this is called as part of a larger flow that already ensures kubeconfig exists, consider adding a comment explaining the dependency.

manifests/post-installation/ovn-template.yaml
```diff
@@ -17,9 +17,7 @@ spec:
         image:
           repository: <OVN_KUBERNETES_IMAGE_REPO>
           tag: <OVN_KUBERNETES_IMAGE_TAG>
-        imagedpf:
-          repository: <OVN_KUBERNETES_UTILS_IMAGE_REPO>
-          tag: <OVN_KUBERNETES_UTILS_IMAGE_TAG>
+        <OVN_KUBERNETES_UTILS_IMAGES>
```
szigmon, line 20: **Verify: Empty placeholder handling**

When `OVN_KUBERNETES_UTILS_IMAGES` is empty (no repo/tag set), this placeholder gets replaced with an empty string. Verify that the resulting YAML is still valid - there may be trailing whitespace or an empty line that could cause issues.

Consider testing with:
```bash
OVN_KUBERNETES_UTILS_IMAGE_REPO="" OVN_KUBERNETES_UTILS_IMAGE_TAG="" make prepare-post-installation
cat manifests/generated/post-installation/ovn-template.yaml
```

szigmon (commented):
Good refactoring to handle multi-line YAML replacements properly. The switch from `sed` to bash parameter expansion is the right approach. A few suggestions below.

scripts/post-install.sh
tsorya: Cause we don't need it here, this part only creates files

wabouhamad (approved):
/lgtm

wabouhamad (commented):
/lgtm

PR #12, opened 2025-12-10
Fix: Remove keepalived from worker nodes to prevent VIP conflicts

itailev:
It seems like in a setup created by the automation, API VIPs can be assigned to all nodes including Worker nodes.
This prevented me to re-deploy a cluster, as the worker nodes of a destroyed cluster was holding the VIP.
The fix:
Adds MachineConfig to remove keepalived static pod manifest from worker nodes. This prevents workers from participating in VRRP and claiming cluster VIPs, which causes installation failures with 'ingress vips already in use' error.

## Summary by CodeRabbit

* **New Features**
  * Automatic removal of keepalived configuration from worker nodes during cluster initialization to ensure proper node setup and prevent stale or conflicting networking configuration. This runs as part of the node boot sequence and confirms removal when present, improving cluster stability and reducing manual cleanup after provisioning for administrators.

### DIFF
--- a/manifests/cluster-installation/99-worker-remove-keepalived.yaml
+++ b/manifests/cluster-installation/99-worker-remove-keepalived.yaml
@@ -1,0 +1,29 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker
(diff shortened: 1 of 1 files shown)


tsorya:
It doesn't seem to be the right way to go, keepalived comes from the pod

tsorya:
Ingress should run on workers, closing this one

PR #15, opened 2025-12-12
Add standalone SR-IOV Device Plugin DaemonSet with hardware-aware supervisor

tsorya:
1. Remove sr-iov operator
2. Small cleanup
3.   Add standalone SR-IOV Device Plugin DaemonSet with hardware-aware supervisor

    Introduce a self-contained SR-IOV Device Plugin deployment that replaces
    dependency on the full SR-IOV Network Operator. This approach is tailored
    for BlueField-3 DPU environments where VF creation timing is unpredictable.

    Supervisor script logic:
    1. PF Discovery: Polls /sys/class/net for Mellanox BF3 devices
       (vendor=15b3, device=a2dc) until found or timeout (600s)

    4. VF Readiness Gate: Blocks plugin startup until ALL VFs are physically
       present in sysfs. This prevents the kubelet from seeing 0 allocatable
       resources during DPU initialization.

    5. Dynamic Configuration: Generates per-node config by introspecting
       actual PCI addresses:
       - VF index 1 → bf3-p0-vfs-mgmt (management traffic)
       - VF index 2+ → bf3-p0-vfs (data plane traffic)

    6. Exec Handoff: Only after hardware validation completes, exec's into
       the actual sriovdp binary with the generated config.

    This pattern ensures:
    - No race conditions between VF creation and device plugin registration
    - Clean separation of management vs data VFs for network isolation
    - Resilient restart behavior (pod restarts on timeout, retries discovery)

## Summary by CodeRabbit

* **New Features**
  * Deploys an SR‑IOV device plugin that auto-discovers PF/VF devices, generates runtime config, and runs on SR‑IOV-capable nodes.

* **Refactor**
  * Removed the SR‑IOV operator subscription and operator config from cluster installation.
  * Removed SR‑IOV network policies from post‑installation manifests.

* **Chores**
  * Streamlined manifest preparation and post‑install flows with batch template substitutions, cleanup of generated values, and added diagnostic logging.


manifests/dpf-installation/sriov-device-plugin.yaml
```diff
@@ -0,0 +1,267 @@
...
+  name: sriov-device-plugin
+  namespace: openshift-sriov-network-operator
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sriov-supervisor-script
+  namespace: openshift-sriov-network-operator
+data:
+  entrypoint.sh: |
```
killianmuldoon, line 49: Could we do this with an init container instead of using a custom script like this?

i.e. init container that has a simple script that waits for the VFs to come up and then the ~normal flow for the SRIOV Device Plugin.

We could inject the init container into the plugin from the SRIOV Network Operator if we wanted as an alternative.

tsorya: We could but it seems more dirty actually as in this case sriov is stuck in init mode.
The problem is that if we want to create an image and not provide a bash script we need to release it somehow as i can't have it in my registry

eelgaev (approved):
Looks good!

PR #25, opened 2026-01-26
fix: add policy to fix hostagent bugs

killianmuldoon:
add a mutatingadmissionpolicy to work around dnspolicy in hostagent pods


killianmuldoon:
This works on my setup - but I haven't gone through the full flow yet.

Validated fully now from scratch

PR #26, opened 2026-01-26
Automate worker and DPU node join

killianmuldoon:


killianmuldoon:
Validated on my setup - this brings up a fully functional DPF cluster which is ready to run full testing against.

tsorya:
Great, thank you.
Nit: Can we add it as a new folder? This logic gonna be part of the operator and will be easier to disable it this way. WDYT?

killianmuldoon:
Can we add it as a new folder? This logic gonna be part of the operator and will be easier to disable it this way.

What do you mean by custom folder? I assume only the DPU part of this will be part of the operator, right?

PR #38, opened 2026-02-02
feat: Add verification for deployment and traffic

killianmuldoon:


szigmon:
**Suggestion**: Auto-detect worker nodes for traffic flow tests                                                                                                
                                                                                                                                                             
When running **traffic-flow-tests.sh**, the script requires **TFT_SERVER_NODE** and **TFT_CLIENT_NODE** to be set manually. 
The current fallback chain  
(HBN_HOSTNAME_NODE* → WORKER_*_NAME) uses BareMetalHost names, not actual Kubernetes node names, which causes pod scheduling failures:                     

0/5 nodes are available: 5 node(s) didn't match Pod's node affinity/selector                                                                                                                                                                                                                                          
Could we add auto-detection from the cluster? Something like:   
```
# Auto-detect worker nodes if not explicitly set                                                                                                           
if [[ -f "${kubeconfig}" ]]; then                                                                                                                          
    workers=$(KUBECONFIG="${kubeconfig}" oc get nodes \                                                                                                    
        -l node-role.kubernetes.io/worker \                                                                                                                
        -l '!node-role.kubernetes.io/master' \                                                                                                             
        -o jsonpath='{.items[*].metadata.name}' 2>/dev/null)                                                                                               
    # Use first two workers as server/client                                                                                                               
fi      
```                                                                                                                                                   
                                                                                                                                                             
This would make the script work out-of-the-box without requiring manual node name configuration.

scripts/verify.sh
```diff
@@ -0,0 +1,209 @@
...
+    local namespace="dpf-operator-system"
+    
+    if ! oc get dpudeployment -n "$namespace" "$name" &>/dev/null; then
+        log "WARN" "DPUDeployment not found, skipping"
+        return 0
+    fi
+    
+    log "INFO" "Waiting for DPUDeployment to be Ready..."
+    
+    for attempt in $(seq 1 "$VERIFY_MAX_RETRIES"); do
```
tsorya: Please use retry 
https://github.com/rh-ecosystem-edge/openshift-dpf/blob/main/scripts/utils.sh#L248

scripts/verify.sh
```diff
@@ -0,0 +1,209 @@
...
+status() {
+    echo "=== Host Cluster Nodes ==="
+    oc get nodes || true
+    
+    echo ""
+    echo "=== DPUDeployment ==="
+    oc get dpudeployment -n dpf-operator-system || true
+    
+    local hosted_kubeconfig="${HOSTED_CLUSTER_NAME}.kubeconfig"
+    if [[ -f "$hosted_kubeconfig" ]]; then
```
tsorya: And if not? not printing anything? Maybe we should update that 0 nodes were found? I believe will be helpful in CI no?

scripts/verify.sh
```diff
@@ -0,0 +1,209 @@
...
+            log "WARN" "DPUCluster kubeconfig not found, skipping"
+            return 0
+        fi
+        oc get secret -n "${CLUSTERS_NAMESPACE}" "${HOSTED_CLUSTER_NAME}-admin-kubeconfig" \
+            -o jsonpath='{.data.kubeconfig}' | base64 -d > "$hosted_kubeconfig"
+    fi
+    
+    log "INFO" "Waiting for $expected_count DPU node(s) to be Ready in DPUCluster..."
+    
+    for attempt in $(seq 1 "$VERIFY_MAX_RETRIES"); do
```
tsorya: Use retry from utils please

scripts/verify.sh
```diff
@@ -0,0 +1,209 @@
...
+    local expected_count="${WORKER_COUNT:-0}"
+    
+    if [[ "$expected_count" -eq 0 ]]; then
+        log "INFO" "WORKER_COUNT=0, skipping DPU node verification"
+        return 0
+    fi
+    
+    local hosted_kubeconfig="${HOSTED_CLUSTER_NAME}.kubeconfig"
+    
+    if [[ ! -f "$hosted_kubeconfig" ]]; then
```
tsorya, line 61: Lets make it a new target in make? Sounds useful to recreate it if required

scripts/verify.sh
```diff
@@ -0,0 +1,209 @@
...
+    log "INFO" "Waiting for $expected_count worker node(s) to be Ready..."
+    
+    for attempt in $(seq 1 "$VERIFY_MAX_RETRIES"); do
+        # Count Ready worker nodes using jq
+        local ready_workers
+        ready_workers=$(oc get nodes -l '!node-role.kubernetes.io/control-plane' -o json | jq '[.items[] | select(.status.conditions[] | select(.type=="Ready" and .status=="True"))] | length')
+        
+        if [[ "$ready_workers" -ge "$expected_count" ]]; then
+            log "INFO" "All $expected_count worker node(s) are Ready"
+            oc get nodes -l node-role.kubernetes.io/worker=
```
tsorya: I am not sure this is the right label as masters with worker label will be shown too and it is confusing as for me as in line 35 you filter out them

killianmuldoon: true - I forgot to update here.

killianmuldoon:
> **Suggestion**: Auto-detect worker nodes for traffic flow tests
> 
> When running **traffic-flow-tests.sh**, the script requires **TFT_SERVER_NODE** and **TFT_CLIENT_NODE** to be set manually. The current fallback chain (HBN_HOSTNAME_NODE* → WORKER_*_NAME) uses BareMetalHost names, not actual Kubernetes node names, which causes pod scheduling failures:
> 
> 0/5 nodes are available: 5 node(s) didn't match Pod's node affinity/selector Could we add auto-detection from the cluster? Something like:
> 
> ```
> # Auto-detect worker nodes if not explicitly set                                                                                                           
> if [[ -f "${kubeconfig}" ]]; then                                                                                                                          
>     workers=$(KUBECONFIG="${kubeconfig}" oc get nodes \                                                                                                    
>         -l node-role.kubernetes.io/worker \                                                                                                                
>         -l '!node-role.kubernetes.io/master' \                                                                                                             
>         -o jsonpath='{.items[*].metadata.name}' 2>/dev/null)                                                                                               
>     # Use first two workers as server/client                                                                                                               
> fi      
> ```
> 
> This would make the script work out-of-the-box without requiring manual node name configuration.

I think this is fine as-is in the sense that we need to set these variables in any case so automatically getting them here doesn't help us really.

scripts/verify.sh
killianmuldoon: What funcitonality do you want in make exactly?

scripts/verify.sh
killianmuldoon: I think we should drop the status function - we aren't using it in this version now and it's not really useful in CI IMO

scripts/verify.sh
tsorya: I would want to be able to download hosted cluster kubeconfig by demand but we can do it later for sure

PR #47, opened 2026-02-05
isolating aicli tokens for users will be able to see the cluster in a…

SharonBX:
aicli token directory (optional). aicli reads ~/.aicli/offlinetoken.txt from HOME.
Default: current user's HOME. Set to use another user's token (e.g. /root/sbx).
When set, scripts validate that AICLI_HOME/.aicli/offlinetoken.txt exists.
OPENSHIFT_PULL_SECRET must be for the same Red Hat account as that token.

## Summary by CodeRabbit

* **New Features**
  * Added `make aicli-list` command to list available clusters
  * Support for AICLI_HOME environment variable to configure alternative user tokens
  * SSH key validation now occurs before cluster creation

* **Documentation**
  * Updated getting-started guide with AICLI_HOME configuration instructions and requirements

### DIFF
--- a/scripts/env.sh
+++ b/scripts/env.sh
@@ -51,6 +51,20 @@ if [ -z "${MAKELEVEL:-}" ]; then
     validate_mtu
 fi
 
+# aicli uses HOME to find ~/.aicli/offlinetoken.txt. Default: $HOME. Override with AICLI_HOME (e.g. in .env).
+# When using AICLI_HOME, OPENSHIFT_PULL_SECRET must be a pull secret for the same Red Hat account as that token.
+AICLI_HOME=${AICLI_HOME:-$HOME}
+if [[ "$AICLI_HOME" != "$HOME" ]] && [[ ! -f "${AICLI_HOME}/.aicli/offlinetoken.txt" ]]; then
+    echo "Error: ${AICLI_HOME}/.aicli/offlinetoken.txt not found." >&2
+    exit 1
+fi
+export HOME="${AICLI_HOME}"
+if ! aicli list clusters &>/dev/null; then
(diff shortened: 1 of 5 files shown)


scripts/env.sh
```diff
@@ -51,6 +51,20 @@ if [ -z "${MAKELEVEL:-}" ]; then
     validate_mtu
 fi
 
+# aicli uses HOME to find ~/.aicli/offlinetoken.txt. Default: $HOME. Override with AICLI_HOME (e.g. in .env).
+# When using AICLI_HOME, OPENSHIFT_PULL_SECRET must be a pull secret for the same Red Hat account as that token.
+AICLI_HOME=${AICLI_HOME:-$HOME}
```
tsorya, line 56: I wonder, maybe we can change aicli repo? And allow folder override in it?

scripts/env.sh
```diff
@@ -51,6 +51,20 @@ if [ -z "${MAKELEVEL:-}" ]; then
...
+if [[ "$AICLI_HOME" != "$HOME" ]] && [[ ! -f "${AICLI_HOME}/.aicli/offlinetoken.txt" ]]; then
+    echo "Error: ${AICLI_HOME}/.aicli/offlinetoken.txt not found." >&2
+    exit 1
+fi
+export HOME="${AICLI_HOME}"
+if ! aicli list clusters &>/dev/null; then
```
tsorya, line 62: Not sure this is the right place for this check. Don't think we need it here, at it to create cluster maybe?

SharonBX: well this check is just for token and make sure we not getting token errors .... no care about the list ...

scripts/env.sh
SharonBX: i looked in aicli code . and i understand wht need to be change ( i think ) 
but i assume it will take much longer then the adaption we do here

tsorya (approved):

PR #55, opened 2026-02-08
Add strict error handling to routing script

tsorya:
Enable set -e for fail-fast behavior and replace permissive '&& echo Success || echo Failure' patterns with direct command calls that let the script exit immediately on failure. Also fix OVN-K subnet detection to find directly connected routes (kernel or DHCP) instead of only DHCP protocol routes.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable pod-to-worker routing with stricter route selection to avoid invalid routes.
  * Routing rules and routes now apply consistently, reducing silent failures and improving network stability.
  * Clearer status messages and a definitive success confirmation during setup.

* **Chores**
  * Service now restarts on failure with a short delay to improve resilience.
  * Added a configuration to prevent unintended interface management.


wabouhamad (approved):
/lgtm

PR #63, opened 2026-02-12
allow user to skip storage operators deploy case we have exsisting st…

SharonBX:
…orageclass

## Summary by CodeRabbit

* **New Features**
  * Added SKIP_DEPLOY_STORAGE flag to allow skipping automatic storage operator deployment and use existing StorageClasses.
  * Added validation to ensure required StorageClasses (via ETCD_STORAGE_CLASS) are present when skipping deployment.

* **Documentation**
  * New "External storage" guide and updated docs/help text explaining SKIP_DEPLOY_STORAGE usage, prerequisites, and mock/test storage examples.


scripts/cluster.sh
```diff
@@ -232,7 +235,12 @@ function start_cluster_installation() {
     wait_for_cluster_status "installed"
     log "INFO" "Cluster installation completed successfully"
     get_kubeconfig
-    if [ "${STORAGE_TYPE}" == "odf" ]; then
+
+    if [ "${SKIP_DEPLOY_STORAGE}" = "true" ]; then
+        log "INFO" "SKIP_DEPLOY_STORAGE=true: validating that required StorageClasses exist (user-provided storage)..."
+        validate_storage_classes_available
```
tsorya, line 241: There can be default storage class and then we don't need to specify one

tsorya: But i believe it is fine for automation

tsorya (approved):

PR #70, opened 2026-02-23
allow user to set static ip for masters VMs

SharonBX:
## Summary by CodeRabbit

* **New Features**
  * Configure VMs with static IPs via VM_STATIC_IP; per-VM static networking (IP, gateway, DNS, MAC) supported and applied before cluster creation
  * VM creation includes an SSH public key by default
  * If a static network file is present, it will be used for VM setup

* **Bug Fixes / Improvements**
  * Improved validation for static networking inputs and clearer handling of the primary network interface

* **Documentation**
  * Added "VMs with static IPs" guide with required variables, optional interface note, and a 3‑VM example

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,28 @@ WORKER_COUNT=2               # Number of physical worker nodes
 make all
 ```
 
+
+### VMs with static IPs
+To assign fixed IPs to cluster VMs (e.g. for predictable addressing or firewall rules), set `VM_STATIC_IP=true` and provide these **required** variables in `.env`:
+
+| Variable     | Description |
+|-------------|----------------------------------------------------------------|
+| `VM_EXT_IPS` | Comma-separated list of static IPs, **one per VM** (must have at least `VM_COUNT` entries). Example: `10.8.2.110,10.8.2.111,10.8.2.112` |
+| `VM_EXT_PL`  | Prefix length for the subnet (e.g. `24` for /24) |
+| `VM_GW`      | Default gateway IP |
+| `VM_DNS`     | DNS server IP(s), comma-separated |
+
+Example for 3 VMs:
+```bash
+VM_STATIC_IP=true
+VM_EXT_IPS=10.8.2.110,10.8.2.111,10.8.2.112
+VM_EXT_PL=24
+VM_GW=10.8.2.1
+VM_DNS=8.8.8.8,4.4.4.4
+```
+
+Optional: `PRIMARY_IFACE` (default `enp1s0`) — interface name used for static config on the VMs.
+
 ## 🛠️ Key Commands
 
 ### Complete Deployment
--- a/scripts/cluster.sh
+++ b/scripts/cluster.sh
@@ -50,12 +50,23 @@ function validate_vips() {
 }
 
 function is_valid_ip() {
-    local ip=$1
-    if [[ $ip =~ ^([0-9]{1,3}\.){3}[0-9]{1,3}$ || $ip =~ ^([0-9a-fA-F]*:[0-9a-fA-F]*){2,}$ ]]; then
-        return 0
-    else
+    local ip="$1"
+
+    if [[ "$ip" =~ [[:space:]] ]]; then
         return 1
     fi
+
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -120,6 +120,28 @@ WORKER_COUNT=2               # Number of physical worker nodes
...
+| `VM_EXT_PL`  | Prefix length for the subnet (e.g. `24` for /24) |
+| `VM_GW`      | Default gateway IP |
+| `VM_DNS`     | DNS server IP(s), comma-separated |
+
+Example for 3 VMs:
+```bash
+VM_STATIC_IP=true
+VM_EXT_IPS=10.8.2.110,10.8.2.111,10.8.2.112
+VM_EXT_PL=24
+VM_GW=10.8.2.1
```
tsorya, line 139: Lets provide different values for dns and gw, just to eliminate confusion

scripts/cluster.sh
```diff
@@ -97,10 +97,100 @@ function set_cluster_mtu() {
...
+          routes:
+            config:
+              - destination: 0.0.0.0/0
+                next-hop-address: ${VM_GW}
+                next-hop-interface: ${PRIMARY_IFACE:-enp1s0}
+EOF
+        done
+
+        return 0
+    fi
```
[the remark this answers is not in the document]
tsorya: @SharonBX seems that rabbit has a point here

scripts/cluster.sh
```diff
@@ -97,10 +97,100 @@ function set_cluster_mtu() {
     fi
     echo "static_network_config:" >> "$STATIC_NET_FILE"
 
+    if [[ "${VM_STATIC_IP:-false}" == "true" ]]; then
```
tsorya: Lets have an early return here, it will make code more readable

scripts/cluster.sh
tsorya: I don't think this the right function name here

scripts/cluster.sh
tsorya: I understand why we need it but maybe lets split it into 2 different flows as it is pretty confusing

scripts/cluster.sh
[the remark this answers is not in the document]
SharonBX: indeed .. will fix

README.md
SharonBX: different values for dns and gw ? not sure i understand

scripts/cluster.sh
```diff
@@ -97,10 +97,100 @@ function set_cluster_mtu() {
...
+        for var in VM_EXT_IPS VM_EXT_PL VM_GW VM_DNS; do
+            if [[ -z "${!var}" ]]; then
+                log "ERROR" "VM_STATIC_IP is enabled but $var is not set"
+                return 1
+            fi
+        done
```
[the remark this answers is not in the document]
SharonBX: added fix lines 108-111

README.md
tsorya: I mean currently we set an example 10.8.2.1 for both, people will think it should be the same ip

scripts/cluster.sh
```diff
@@ -97,10 +97,105 @@ function set_cluster_mtu() {
...
+                   prefix-length: ${VM_EXT_PL}
+          dns-resolver:
+            config:
+              server:
+${DNS_SERVERS_YAML}
+          routes:
+            config:
+              - destination: 0.0.0.0/0
+                next-hop-address: ${VM_GW}
+                next-hop-interface: ${PRIMARY_IFACE:-enp1s0}
```
tsorya: Lets handle default in env.sh please

scripts/cluster.sh
```diff
@@ -97,10 +97,105 @@ function set_cluster_mtu() {
     fi
     echo "static_network_config:" >> "$STATIC_NET_FILE"
 
+    if [[ "${VM_STATIC_IP}" == "true" ]]; then
```
tsorya: Lets move all validations to a function.

tsorya (approved):

PR #80, opened 2026-03-01
NVIDIA-507: Update the provisioner image repository and chart URL, and set their versions to `latest`.

linoyaslan:
Update the provisioner `DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REPO` and `DPF_HCP_PROVISIONER_OPERATOR_CHART_URL` to the repositories created by Konflux + and set their versions `DPF_HCP_PROVISIONER_OPERATOR_VERSION` and  `DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG` to `latest`.

## Summary by CodeRabbit

* **Chores**
  * Updated default DPF HCP Provisioner Operator configuration to point at the new OCI chart location and image repository, and to use the "latest" image tag by default.
  * Default chart version is now empty so tooling will omit an explicit --version when not set.
* **Bug Fixes**
  * Helm upgrade now omits the version flag when no version is specified, avoiding unintended version pinning.


scripts/env.sh
```diff
@@ -253,11 +253,11 @@ SANITY_TESTS_PING_COUNT=${SANITY_TESTS_PING_COUNT:-"20"}
...
 # DPF HCP Provisioner Operator Configuration
-DPF_HCP_PROVISIONER_OPERATOR_CHART_URL=${DPF_HCP_PROVISIONER_OPERATOR_CHART_URL:-"oci://quay.io/lhadad/charts/dpf-hcp-provisioner-operator"}
+DPF_HCP_PROVISIONER_OPERATOR_CHART_URL=${DPF_HCP_PROVISIONER_OPERATOR_CHART_URL:-"oci://quay.io/redhat-user-workloads/dpf-hcp-provisioner-tenant/chart"}
 DPF_HCP_PROVISIONER_OPERATOR_NAMESPACE=${DPF_HCP_PROVISIONER_OPERATOR_NAMESPACE:-"dpf-hcp-provisioner-system"}
-DPF_HCP_PROVISIONER_OPERATOR_VERSION=${DPF_HCP_PROVISIONER_OPERATOR_VERSION:-"0.1.2"}
-DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REPO=${DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REPO:-"quay.io/lhadad/dpf-hcp-provisioner-operator"}
-DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG=${DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG:-"v0.1.2"}
+DPF_HCP_PROVISIONER_OPERATOR_VERSION=${DPF_HCP_PROVISIONER_OPERATOR_VERSION:-"latest"}
+DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REPO=${DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REPO:-"quay.io/redhat-user-workloads/dpf-hcp-provisioner-tenant/dpf-hcp-provisioner"}
+DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG=${DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG:-"latest"}
```
linoyaslan, line 262: @tsorya, I’d appreciate your opinion here.
Technically, Helm doesn’t allow `latest` as a chart version since it’s not a valid semver. What actually happens is that when we set it to `latest`, the automation ends up passing an empty `--version` flag, meaning we don’t explicitly set a version, and Helm then pulls the highest available semver from the OCI registry. Claude suggested that keeping `latest` as the default in `env.sh` is clearer for readers, since it better communicates the intent ("use the latest version") compared to leaving it empty.
FYI we do have a `latest` tag in quay.io (I added it in the tekton files), but pulling with the `latest` tag doesn’t actually work and results in an error due to the explanation above.

thoughts?

linoyaslan: for example:
```
➜  helm pull oci://quay.io/redhat-user-workloads/dpf-hcp-provisioner-tenant/chart:latest                  
Error: improper constraint: latest
➜  helm pull oci://quay.io/redhat-user-workloads/dpf-hcp-provisioner-tenant/chart       
Pulled: quay.io/redhat-user-workloads/dpf-hcp-provisioner-tenant/chart:0.1.48_c34e3a1
Digest: sha256:dd6dd13852c38d1035ff5d00f000dda9240a5d442a493adb7590ddc3bb5d3993
quay.io/redhat-user-workloads/dpf-hcp-provisioner-tenant/chart:0.1.48_c34e3a1 contains an underscore.

OCI artifact references (e.g. tags) do not support the plus sign (+). To support
storing semantic versions, Helm adopts the convention of changing plus (+) to
an underscore (_) in chart version tags when pushing to a registry and back to
a plus (+) when pulling from a registry.

```

tsorya (approved):

PR #81, opened 2026-03-02
Add worker static IP manifests and update worker scripts

SharonBX:
## Summary by CodeRabbit

* **New Features**
  * Added support for static IP configuration on worker nodes with per-worker customization.

* **Documentation**
  * Updated provisioning guides with static IP setup instructions and configuration examples.

* **Chores**
  * Added new environment variable for static IP provisioning control.


manifests/worker-provisioning/bmc-secret-static-ip.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: <WORKER_NAME>-network-config
+  namespace: openshift-machine-api
+type: Opaque
+stringData:
+  nmstate: |
+    interfaces:
+      - name: br-dpu
```
tsorya, line 10: I am not sure regarding this one, how it will work together with machine config script where we set br-dpu?

manifests/worker-provisioning/baremetalhost-static-ip.yaml
```diff
@@ -0,0 +1,23 @@
+---
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
```
tsorya, line 3: can we have have a single template for both cases?

manifests/worker-provisioning/bmc-secret-static-ip.yaml
SharonBX: machine config will apply regular but  apply-nmstate-bridge will skip because br-dpu already exists

manifests/worker-provisioning/baremetalhost-static-ip.yaml
SharonBX: here we have extra spec key preprovisioningNetworkDataName: <WORKER_NAME>-network-config
its ether use two template as now or handle it in worker.sh ...

manifests/worker-provisioning/bmc-secret-static-ip.yaml
tsorya: The question is what want here, we want to be sure that apply-nmstate-bridge works in case of static ip or we want to drop this script

tsorya: Didn't we agree to change it?

PR #83, opened 2026-03-03
Bump DPF to 26.4 pre-release

killianmuldoon:
This commit bumps versions to use a pre-release build of DPF 26.4 and a newer version of the OVN Kubernetes helm chart.

It adds a number of workarounds to enable DPF 26.4:

- fixes values on OVN Kubernetes to disable features which don't work with the OCP 4.21 version of OVN Kubernetes
- Uses MutatingWebhookConfiguration for ovn-kubernetes device inejction and drops MutatingAdmissionPolicy feature enablement
- Sets hostAgentDNSPolicy in DPFOperatorConfig and drops MutatingWebhookPolicy workaround
- Adds missing RBAC for controllers - for SCC and to fix a bug in the provisioning controller
- Adds a workaround for DNS resolution in the BFB registry pod

## Summary by CodeRabbit

* **New Features**
  * Automatic registry-address injection for provisioning and new host-agent DNS policy option
  * RBAC/SCC bindings added to support controller operations

* **Chores**
  * Updated OpenShift, DPF and OVN component versions and chart/image defaults
  * Added OVN helm configuration flags and OCI-based chart hosting

* **Refactor**
  * Replaced legacy admission-policy DNS mutation with webhook-based injector flow and node-label targeting
  * Removed strict DPU_INTERFACE requirement and legacy hostagent mutation


scripts/worker.sh
```diff
@@ -24,6 +24,9 @@ provision_all_workers() {
     # Apply short worker hostnames MachineConfig if enabled
     apply_short_worker_hostnames
 
+    # Apply custom node labels MachineConfig if configured
+    apply_worker_node_labels
```
vasrem, line 28: what if the workers dont have a DPU?

killianmuldoon: This is automation for our CI - the workers always have a DPU.

scripts/manifests.sh
```diff
@@ -332,11 +332,22 @@ prepare_dpf_manifests() {
...
     
     # Process dpfoperatorconfig.yaml
+    # Get node IP for BFB registry address (workaround for hostagent DNSPolicy:Default)
+    local node_ip
+    node_ip=$(oc get nodes -o jsonpath='{.items[0].status.addresses[?(@.type=="InternalIP")].address}')
+    if [ -z "$node_ip" ]; then
+        log "ERROR" "Failed to get node InternalIP for BFB registry address"
+        return 1
+    fi
+    local bfb_registry_address="http://${node_ip}:30082"
```
tsorya, line 342: What this registry is used for? There is a pod running? Why we use host[0]?

scripts/worker.sh
tsorya: Yes but i believe we need to understand how we should handle it in real env

killianmuldoon: Sure - but out of scope in this PR.

scripts/manifests.sh
killianmuldoon: In 26.4 all BFBs are pulled from this registry - so the flow goes:
1) BFB controller downloads BFB
2) BFB controller pushes BFB to registry
3) HostAgent pulls BFB from registry


This is a workaround until we can incorporate a bug fix for communication between the host agent and the registry - we changed the dnsPolicy of the host agent which means it can't use the Kubernetes service to contact the registry. This will be fixed in 26.4 and users will not have to set it.

We need any host - host[0] in this case because it's a nodePort service.

scripts/worker.sh
tsorya: @killianmuldoon what do you think if we will change the logic here, can we add systemd service that will set the label if br-dpu bridge exists?

manifests/worker-provisioning/99-worker-node-labels.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: machineconfiguration.openshift.io/v1
```
tsorya, line 1: Do we need new machine config? We already have one why not to use the same?

killianmuldoon: IMO it's cleaner to have these separated out for now. Is this a blocker for 26.4 integration?

tsorya (approved):

PR #89, opened 2026-03-10
fix: only include static net file when VM_STATIC_IP is true

killianmuldoon:
## Summary by CodeRabbit

* **Refactor**
  * Reorganized cluster creation script to use a centralized argument array approach, consolidating how command arguments are constructed and passed to the cluster creation utility while maintaining identical functionality and behavior.

### DIFF
--- a/scripts/cluster.sh
+++ b/scripts/cluster.sh
@@ -263,32 +263,36 @@ function check_create_cluster() {
         
         ensure_ssh_key_in_home || return 1
         
+        local aicli_args=()
         if [ "$VM_COUNT" -eq 1 ]; then
             log "INFO" "Creating single-node cluster..."
-            aicli create cluster \
-                -P openshift_version="${OPENSHIFT_VERSION}" \
-                -P base_dns_domain="${BASE_DOMAIN}" \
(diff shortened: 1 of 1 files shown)


scripts/cluster.sh
```diff
@@ -263,32 +263,36 @@ function check_create_cluster() {
...
+                -P openshift_version="${OPENSHIFT_VERSION}"
+                -P base_dns_domain="${BASE_DOMAIN}"
+                -P api_vips="${API_VIPS}"
+                -P pull_secret="${OPENSHIFT_PULL_SECRET}"
+                -P public_key="${SSH_KEY}"
+                -P ingress_vips="${INGRESS_VIPS}"
+            )
         fi
+        if [[ "${VM_STATIC_IP}" == "true" ]]; then
+            aicli_args+=(--paramfile "${STATIC_NET_FILE}")
```
SharonBX, line 292: user MTU is pass through the static net file for nmstate proccess .  
looks like its been broken ...

killianmuldoon: Yeah - I'm still wondering what the source of the issue is here. Did you manage to test with OCP 4.21?

PR #90, opened 2026-03-10
Drop NFS requirement for multi-node clusters

killianmuldoon:
## Summary by CodeRabbit

* **Chores**
  * Removed built-in NFS server support and related deployment/configuration options, environment defaults, and manifests.
  * Updated documentation and manifests to simplify storage to 3 volumes for etcd (no NFS PV).

* **Bug Fixes**
  * Improved OpenShift compatibility by applying security-context patches to provisioning and registry components during installation.


scripts/dpf.sh
```diff
@@ -97,6 +97,28 @@ function deploy_metallb() {
     log [INFO] "Note: IPAddressPool and L2Advertisement will be managed by dpf-hcp-provisioner-operator"
 }
 
+function patch_provisioning_init_container() {
```
tsorya, line 100: Will we have a fix for it in GA?

scripts/post-install.sh
```diff
@@ -331,6 +331,36 @@ function apply_post_installation() {
...
+    patch_bfb_registry
+}
+
+function patch_bfb_registry() {
+    # The bfb-registry is a standalone pod (not a Deployment) that serves BFB files via nginx.
+    # On OpenShift, SELinux blocks non-privileged containers from reading HostPath files written
+    # by other pods. The bfb-registry needs privileged: true to read files in /bfb/bfcfg/.
+    # Since standalone pod securityContext is immutable, we must replace the pod.
+    log [INFO] "Waiting for bfb-registry pod..."
+    if ! retry 30 10 oc get pod -n dpf-operator-system bfb-registry &>/dev/null; then
```
tsorya, line 360: Ill ask here something, why it is not a deployment? and another one ? will it be fixed in GA?

scripts/dpf.sh
killianmuldoon: Yeah - we'll have a fix for this one.

scripts/post-install.sh
killianmuldoon: This will be fixed in GA for sure. As to why this isn't a deployment I'm not certain, but e.g. the DMS pods are deployed in the same way with this controller as the owner.

tsorya (approved):

PR #96, opened 2026-03-16
Automate OVN-Kubernetes image tag

eelgaev:
This PR simply selects the right tag for the OVN-K image from the OCP release payload for aarch64.

## Summary by CodeRabbit

* **Chores**
  * OCP release image default now derives from the configured OpenShift version instead of a fixed image.
  * OVN-Kubernetes image defaults cleared (no predefined repository or tag).
  * Optional automatic resolution of the OVN-Kubernetes image from the OpenShift release payload is added when release tooling is available.


scripts/env.sh
```diff
@@ -139,6 +139,18 @@ if [[ "${BASH_SOURCE[0]}" != "${0}" ]]; then
...
 
+    # Auto-resolve OVN-Kubernetes image from the aarch64 OCP release payload
+    _ocp_version="${OCP_RELEASE_IMAGE##*:}"    # e.g. 4.21.0-multi
+    _ocp_version="${_ocp_version%%-*}"          # e.g. 4.21.0
+    if command -v oc &>/dev/null; then
+        _ovnk_full=$(oc adm release info --image-for=ovn-kubernetes \
+            "quay.io/openshift-release-dev/ocp-release:${_ocp_version}-aarch64" 2>/dev/null || true)
+        if [ -n "$_ovnk_full" ]; then
+            OVN_KUBERNETES_IMAGE_REPO="${_ovnk_full%@*}@sha256"
+            OVN_KUBERNETES_IMAGE_TAG="${_ovnk_full##*sha256:}"
```
killianmuldoon, line 149: We should be able to overwrite this image if required though, right? So we can use different versions of OVN Kubernetes if required.

eelgaev: Uhmm.
Let me look into that, my bad.

eelgaev: Fixed that I believe, line 144

ci/env.defaults
```diff
@@ -26,8 +26,8 @@ DPF_HELM_REPO_URL=${DPF_HELM_REPO_URL:-oci://ghcr.io/killianmuldoon/doca-platfor
 OVN_CHART_URL=${OVN_CHART_URL:-oci://ghcr.io/mellanox/charts}
 OVN_TEMPLATE_CHART_URL=${OVN_TEMPLATE_CHART_URL:-oci://ghcr.io/mellanox/charts}
 OVN_CHART_VERSION=${OVN_CHART_VERSION:-v25.7.1-79ed186}
-OVN_KUBERNETES_IMAGE_REPO=${OVN_KUBERNETES_IMAGE_REPO:-quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256}
-OVN_KUBERNETES_IMAGE_TAG=${OVN_KUBERNETES_IMAGE_TAG:-0f8e255375dacfb26a0b1726a7e18ec9e67540cd7e0950bf0a8dcea0ba991cb9}
+OVN_KUBERNETES_IMAGE_REPO=${OVN_KUBERNETES_IMAGE_REPO:-}
+OVN_KUBERNETES_IMAGE_TAG=${OVN_KUBERNETES_IMAGE_TAG:-}
```
killianmuldoon: We should remove these from defaults if the defaulting is done elsewhere IMO

killianmuldoon (commented):
One last finding - looks good otherwise.

tsorya (approved):

PR #98, opened 2026-03-18
Add worker static IP manifests and update worker scripts

SharonBX:
## Summary by CodeRabbit

* **New Features**
  * Per-worker static IP provisioning with automated manifest generation and validation.

* **Documentation**
  * Added "Workers with static IPs" guide with examples and deployment behavior notes.

* **Configuration**
  * New WORKER_STATIC_IP environment variable and template declarations.

* **Provisioning**
  * Enhanced bridge setup to generate/apply DHCP or STATIC NMState configs with improved logging and error handling.

* **Manifests**
  * Added static-IP BareMetalHost and Secret manifests; updated service unit ordering and restart behavior for bridge setup.


configuration_templates/apply-nmstate-bridge.sh
```diff
@@ -1,86 +1,112 @@
...
-rm "$TMP_FILE"
+    cat <<EOF > /tmp/br-dpu-config.yml
+interfaces:
+  - name: $BRIDGE_NAME
+    type: linux-bridge
+    state: up
+    mac-address: $MAC_ADDR
+    mtu: $TARGET_MTU
+    ipv4:
+      enabled: true
```
tsorya, line 80: and what about ipv6? Do we want to add support later?

configuration_templates/apply-nmstate-bridge.sh
```diff
@@ -1,86 +1,112 @@
 #!/bin/bash
+set -e
 
-DEFAULT_MTU=1500
-NODES_MTU=${1:-$DEFAULT_MTU}
+BRIDGE_NAME="br-dpu"
+IP_HINT_FILE="/run/nodeip-configuration/primary-ip"
+TARGET_MTU="${1:-1500}"
```
tsorya, line 6: Maybe lets take mtu from interface ? always?

SharonBX: but in dhcp mode this is the 1st time we really use etu env var . 
in static we use user mtu from IPA to CoreOS

configuration_templates/apply-nmstate-bridge.sh
SharonBX: we can add support easily but needs our lab to support it first .

PR #105, opened 2026-03-20
fix: bump openshift version in defaults

killianmuldoon:
bump openshift version to one supported by the HCP provisioner

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift version to 4.21.1

### DIFF
--- a/ci/env.defaults
+++ b/ci/env.defaults
@@ -6,7 +6,7 @@
 # Syntax: ${VAR:-default} means "use default if VAR is unset or empty"
 
 # Cluster Configuration
-OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-4.21.0}
+OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-4.21.1}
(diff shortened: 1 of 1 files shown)


ci/env.defaults
```diff
@@ -6,7 +6,7 @@
 # Syntax: ${VAR:-default} means "use default if VAR is unset or empty"
 
 # Cluster Configuration
-OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-4.21.0}
+OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-4.21.1}
```
tsorya, line 9: We should move to 4.22

killianmuldoon: Do you have the full set of variables required to work with 4.22?

killianmuldoon: I see this was already done in https://github.com/rh-ecosystem-edge/openshift-dpf/pull/108 will close

PR #106, opened 2026-03-20
Skip NMState paramfile for assisted installer when MTU is 1500 and no static IP

tsorya:
When NODES_MTU is 1500 (or unset) and VM_STATIC_IP is not enabled, there is no need to send NMState configuration to the assisted installer. The default DHCP behavior works without it. The paramfile is still generated and sent for MTU 9000 (to set jumbo frames) or when static IP configuration is required.

Made-with: Cursor

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized cluster creation process by eliminating unnecessary network configuration generation under specific scenarios.
  * Enhanced parameter handling to reduce redundant operations during cluster initialization, improving overall efficiency.

### DIFF
--- a/scripts/cluster.sh
+++ b/scripts/cluster.sh
@@ -169,6 +169,12 @@ function set_node_nmstate() {
     if [ -f "$STATIC_NET_FILE" ]; then
         rm "$STATIC_NET_FILE"
     fi
+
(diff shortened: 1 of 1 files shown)


wabouhamad (approved):
/lgtm

PR #107, opened 2026-03-23
No need to provide API and IGRESS vips in case of sno

tsorya:
## Summary by CodeRabbit

* **Bug Fixes**
  * Made API and Ingress VIP validation conditional—now only required for multi-node deployments. Single-node cluster configurations can proceed without these settings, reducing unnecessary configuration requirements.


wabouhamad (approved):
/lgtm

PR #114, opened 2026-03-31
Add --disable-openapi-validation to helm installs and exclude Package…

tsorya:
…Manifest from ArgoCD

Adds --disable-openapi-validation flag to all three helm upgrade --install commands (dpf-hcp-provisioner-operator, maintenance-operator, dpf-operator) to avoid OpenAPI validation failures. Configures ArgoCD resource exclusions for PackageManifest resources to prevent sync issues.

Made-with: Cursor

## Release Notes

* **Chores**
  * Updated ArgoCD resource management configuration to exclude specific operator resources from synchronization
  * Disabled OpenAPI validation during Helm deployments for improved deployment compatibility

### DIFF
--- a/manifests/gitops-operator/argocd.yaml
+++ b/manifests/gitops-operator/argocd.yaml
@@ -5,6 +5,14 @@ metadata:
   name: argocd
   namespace: dpf-operator-system
 spec:
+  resourceExclusions: |
+    - apiGroups:
+      - packages.operators.coreos.com
+      kinds:
+      - PackageManifest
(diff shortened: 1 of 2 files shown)


killianmuldoon:
Why do we want do disable openAPI validation?

tsorya:
While working on OCP 4.22 i see that there are some broken apis, not interesting for our case. Hopefully will be fixed

eelgaev (approved):
LGTM

PR #120, opened 2026-04-07
fix: set LINK_TYPE_P1/P2=ETH in DPUFlavor nvconfig

itailev:
Explicitly set LINK_TYPE_P1=ETH and LINK_TYPE_P2=ETH in both DPUFlavor manifests (1500 and 9000 MTU variants). IB link type is not supported in DPF and leaving this unset can result in the wrong default.

## Summary by CodeRabbit

* **Configuration**
  * Added support for configurable Ethernet link types on supported device flavors, enabling more flexible network configuration options.


tsorya:
Fixed by https://github.com/rh-ecosystem-edge/openshift-dpf/pull/119

PR #127, opened 2026-04-10
Fix VM deletion nuking all VMs when prefix is empty

tsorya:
When VM_PREFIX or VM_WORKER_PREFIX was empty, grep "" matched every line from virsh list --all, destroying all VMs on the hypervisor.

- Guard _delete_vms_by_prefix against empty prefix (skip with warning)
- Use anchored grep on VM name column only (grep "^${prefix}")
- Skip worker VM deletion when VM_WORKER_PREFIX is not set
- Skip VM deletion in clean_all when VM_PREFIX is empty

Made-with: Cursor

## Release Notes

* **Bug Fixes**
  * Safer virtual machine cleanup: deletion is skipped when required configuration is missing to avoid unintended removals.
  * More precise VM matching to reduce false-positive deletions during cleanup.

* **Improvements**
  * Better logging when parameters are absent so skipped cleanup actions are visible to operators.
  * Pipeline robustness: cleanup steps no longer abort on intermediate command failures.

### DIFF
--- a/scripts/cluster.sh
+++ b/scripts/cluster.sh
@@ -465,8 +465,12 @@ function clean_all() {
     delete_cluster
     
     # Delete VMs
-    log "INFO" "Deleting VMs with prefix $VM_PREFIX..."
-    scripts/vm.sh delete || true
+    if [ -n "${VM_PREFIX}" ]; then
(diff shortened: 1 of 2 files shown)


wabouhamad (approved):
/lgtm

PR #133, opened 2026-04-13
Move static net files to generated dir instead of configuration_templates

tsorya:
Move static net files to generated dir instead of configuration_templates

The STATIC_NET_FILE and WORKER_STATIC_NET_FILE were defaulting to ./configuration_templates/ which is not part of the project structure, not cleaned up by clean_resources(), and not in .gitignore. Move them to $GENERATED_DIR where all other generated artifacts live.

Made-with: Cursor

## Summary by CodeRabbit

* **Chores**
  * Updated default file path configuration for static network and worker state files to use generated directories instead of template locations for improved consistency.
  * Enhanced cluster setup and deployment scripts with improved directory management, ensuring parent directories are automatically created before file operations.


eelgaev (approved):
LGTM and tested right now, it's working

PR #134, opened 2026-04-13
Add deploy-lvms to make all and make the target functional

tsorya:
The deploy-lvms Makefile target was a no-op that only printed an info message. Wire it to actually call deploy_lvm, add guards for SKIP_DEPLOY_STORAGE and STORAGE_TYPE!=lvm, and include it in the _all target so LVM is deployed even when re-running against an already installed cluster.

Made-with: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LVMS deployment execution via the `deploy-lvms` make target to properly deploy the LVM Storage operator instead of only displaying informational messages.
  * Added configuration controls to skip LVM Storage deployment when appropriate based on storage type settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,7 +211,7 @@ deploy-odf:
 	@$(CLUSTER_SCRIPT) deploy-odf
 
 deploy-lvms:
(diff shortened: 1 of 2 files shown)


eelgaev (approved):
Looks good

PR #137, opened 2026-04-14
Disable DPU health check and clear OVN utils image defaults

tsorya:
- Add dpuHealthCheck with renewInterval: 0 to OVN configuration to disable the DPU health check lease mechanism
- Clear default values for OVN_KUBERNETES_UTILS_IMAGE_REPO and OVN_KUBERNETES_UTILS_IMAGE_TAG so the imagedpf section is omitted from ovn-template.yaml unless explicitly set

Made-with: Cursor


manifests/post-installation/ovn-configuration.yaml
```diff
@@ -16,6 +16,8 @@ spec:
         podNetwork: 10.128.0.0/14/23
         serviceNetwork: 172.30.0.0/16
         mtu: <NODES_MTU>
+        dpuHealthCheck:
+          renewInterval: 0
```
killianmuldoon, line 20: Trying this I'm getting an error:

```
invalid lease config: renewInterval (0) and leaseDuration (40) must be > 0, and leaseDuration must be greater than renewInterval
````

tsorya:
we can't disable healthcheck as it is enabled on host

PR #154, opened 2026-04-22
Igal/fix nmstate bridge ordering

tsorya:
fix: order nmstate-bridge.service before crio and kubelet
During boot, nmstate-bridge.service creates the br-dpu bridge and moves
eno12399 under it. This briefly removes the node IP from the system while
br-dpu acquires a new DHCP lease. If crio starts during this window, it
fails to bind its streaming server to the node IP (EADDRNOTAVAIL), which
cascades into a kubelet dependency failure that doesn't auto-recover.

Adding Before=crio.service kubelet.service ensures the bridge is fully
configured and the IP is available before either service starts.

rename worker-nmstate-bridge to dpu-worker-configuration
Rename the MachineConfig and its manifest file to better reflect
its purpose as the DPU worker node configuration.

Made-with: Cursor

## Summary by CodeRabbit

* **Chores**
  * Renamed the DPU worker configuration artifact and updated manifest handling to use the new configuration name.

* **Bug Fixes**
  * Ensures DPU init runs before the container runtime and kubelet, and adds an explicit startup timeout.
  * Improves bridge IP detection and interface selection by waiting for the bridge to acquire the node IP for more reliable network setup.

### DIFF
--- a/manifests/cluster-installation/99-dpu-worker-configuration.yaml
+++ b/manifests/cluster-installation/99-dpu-worker-configuration.yaml
@@ -3,7 +3,7 @@ kind: MachineConfig
 metadata:
   labels:
     machineconfiguration.openshift.io/role: worker
-  name: worker-nmstate-bridge
+  name: dpu-worker-configuration
 spec:
   config:
     ignition:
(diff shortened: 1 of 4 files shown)


wabouhamad (approved):
/lgtm

PR #156, opened 2026-04-24
fix: drop unnecessary labels from argoCD object

killianmuldoon:


manifests/gitops-operator/argocd.yaml
```diff
@@ -27,20 +27,6 @@ spec:
...
-  controller:
-    labels:
-      ovn.dpu.nvidia.com/skip-injection: ""
-
-  repo:
-    labels:
-      ovn.dpu.nvidia.com/skip-injection: ""
-
-  redis:
-    labels:
```
killianmuldoon, line 42: This field is no longer defined in newer versions of the ArgoCD CRD so it's causing errors. We can remove them without issue with the new injector.

tsorya (approved):

PR #163, opened 2026-05-04
Add worker-dpu MachineConfigPool support for DPU workers

SharonBX:
This commit adds support for provisioning DPU workers into a dedicated worker-dpu MachineConfigPool with DPU-specific configuration.

New features:
- worker-dpu MachineConfigPool definition
- Worker MCP updated to exclude worker-dpu nodes
- DPU detection via BF3 PCI device (vendor 15b3, devices a2d6/a2dc)
- OVS services disabled on DPU nodes via systemd dropins

Key files:
- manifests/cluster-installation/99-worker-dpu-mcp.yaml: MCP definition
- manifests/cluster-installation/99-worker-mcp-update.yaml: Worker MCP exclusion
- manifests/cluster-installation/99-worker-dpu-mc-list.yaml: DPU detection and service configuration
- manifests/worker-provisioning/machineset-dpu.yaml: Shared MachineSet template
- manifests/worker-provisioning/baremetalhost-dpu.yaml: BMH template for DPU workers
- scripts/worker.sh: Provisioning logic for DPU workers
- Makefile: Added WORKER_n_DPU config support

Usage:
  WORKER_1_DPU=true in .env to provision worker as DPU-capable node

## Summary by CodeRabbit

* **New Features**
  * Add end-to-end support for provisioning and managing DPU (Data Processing Unit) workers: DPU-aware machine sets, a dedicated DPU worker pool, and automated labeling for DPU nodes.
  * Conditional provisioning flows that generate appropriate host resources and apply DPU-specific configuration where applicable.
  * Runtime detection for DPU-capable hardware to mark nodes for DPU handling.

* **Documentation**
  * Help output and environment template updated to expose DPU-related configuration variables.

PR #166, opened 2026-05-07
feat: add LIBVIRT_HOST support for remote VM hosting

eelgaev:
Allow running the automation locally while hosting VMs on a remote libvirt host over SSH. When LIBVIRT_HOST is set (e.g., root@192.168.1.100), all virsh/virt-install commands use qemu+ssh:// URI, ISOs are downloaded directly on the remote host via curl, and bridge setup runs remotely. When unset, behavior is unchanged (local qemu:///system).

This PR takes us one step closesr to a full agentic workflow.

## Summary by CodeRabbit

* **New Features**
  * Support for a configurable remote libvirt host (LIBVIRT_HOST) to provision and manage VMs over SSH, including remote ISO placement, network bridge management, and remote URI selection.

* **Documentation**
  * Added user guide section describing the remote libvirt host workflow, prerequisites (SSH keys, libvirtd), and behavior changes for paths/operations when remote.

* **Chores**
  * Updated help text and environment templates to expose the new LIBVIRT_HOST setting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/openshift-dpf/pull/166)


eelgaev:
This is work in progress.

eelgaev:
## Testing update — remote libvirt with SNO

Tested the remote libvirt flow end-to-end on `nvd-srv-06` with a single-node cluster. Found and fixed two issues:

### Fixes pushed (commit 7880286):

**1. ANSI color codes in ISO URL**
`aicli info iso <cluster> -s` emits terminal color escape sequences (`\x1b[0;36m`) that get embedded in the URL string. When passed via SSH to `curl` on the remote host, this corrupts the URL. Fixed by stripping ANSI codes with `sed`.

**2. Missing `curl -g` (globoff) flag**
JWT tokens in Assisted Installer ISO URLs can contain bracket characters that `curl` interprets as URL globbing ranges, causing `bad range in URL` errors. Added `-g` flag to disable globbing.

Both issues only manifest with remote libvirt since local downloads use `aicli download iso` directly.

### Test environment
- Remote host: nvd-srv-06 (x86_64, QEMU 9.1.0, libvirt 10.10.0)
- Local host: aarch64 Fedora (cross-arch remote libvirt)
- Cluster: SNO, OCP 4.22.0-rc.1
- VM on bridge `br0`, DHCP on lab network
- Result: ✅ Cluster installed successfully

— Lyra

eelgaev:
### CodeRabbit review follow-up

Went through the CodeRabbit comments:

**1. ARP/ip-neigh fallback is local-only in remote mode** — ✅ Valid, good catch. Fixed in commit 8df9490. The `get_vm_ip_by_mac()` now SSH's to `$LIBVIRT_HOST` for ARP lookups when in remote mode. We hit this exact issue during testing — had to manually SSH to the hypervisor to find the VM IP.

**2. Add curl timeout for remote ISO download** — Reasonable but not critical. The SSH session itself will timeout eventually, and adding `--max-time 3600` for an ISO download is arbitrary. Skipping for now.

**3. Quote injection in SSH command** — Technically correct but practically irrelevant. `download_path`, `cluster_name`, and `iso_url` are all controlled by our own env vars, not user input. The heredoc approach adds complexity for a non-existent attack surface. Skipping.

— Lyra

scripts/cluster.sh
```diff
@@ -714,11 +711,18 @@ function get_iso() {
...
     fi
 
-    mkdir -p "${download_path}" || true
-
-    if ! aicli download iso "${cluster_name}" -p "${download_path}"; then
-        log "ERROR" "Failed to download ISO for cluster ${cluster_name}"
-        return 1
+    if is_remote_libvirt; then
+        log "INFO" "Downloading ISO directly on remote host ${LIBVIRT_HOST}..."
+        if ! ssh "${LIBVIRT_HOST}" "mkdir -p '${download_path}' && curl -fLo '${download_path}/${cluster_name}.iso' '${iso_url}'"; then
```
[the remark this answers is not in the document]
eelgaev: The SSH session itself provides a natural timeout boundary. Adding `--max-time 3600` is arbitrary — ISOs vary in size, and network speeds differ across environments. If someone's ISO download is hanging, the problem is connectivity, not a missing curl flag. Skipping for now.

— Lyra

scripts/cluster.sh
```diff
@@ -714,11 +711,18 @@ function get_iso() {
...
 
-    mkdir -p "${download_path}" || true
-
-    if ! aicli download iso "${cluster_name}" -p "${download_path}"; then
-        log "ERROR" "Failed to download ISO for cluster ${cluster_name}"
-        return 1
+    if is_remote_libvirt; then
+        log "INFO" "Downloading ISO directly on remote host ${LIBVIRT_HOST}..."
+        if ! ssh "${LIBVIRT_HOST}" "mkdir -p '${download_path}' && curl -gfLo '${download_path}/${cluster_name}.iso' '${iso_url}'"; then
+            log "ERROR" "Failed to download ISO on remote host ${LIBVIRT_HOST}"
```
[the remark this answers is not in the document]
eelgaev: Technically correct but not a real risk here. `download_path`, `cluster_name`, and `iso_url` are all derived from our own env vars — not user-controlled input. The heredoc approach adds complexity for a non-existent attack surface in an internal automation tool. Keeping it simple.

— Lyra

scripts/cluster.sh
```diff
@@ -15,9 +15,6 @@
 set -e
 set -o pipefail
 
-# Source environment variables
-source "$(dirname "$0")/env.sh"
```
tsorya, line 19: why?

scripts/cluster.sh
```diff
@@ -700,7 +697,7 @@ function get_iso() {
     fi
 
     log "INFO" "Getting ISO URL..."
-    local iso_url="$(aicli info iso "${cluster_name}" -s)"
+    local iso_url="$(aicli info iso "${cluster_name}" -s | sed 's/\x1b\[[0-9;]*m//g')"
```
tsorya, line 700: Explain please

scripts/update-etc-hosts.sh
```diff
@@ -43,25 +43,32 @@ get_vm_ip() {
...
     if [ -z "$vm_mac" ]; then
         echo "Error: Could not retrieve MAC address for VM: $vm_name" >&2
         return 1
     fi
 
     # Use arp or ip neigh to find the IP address based on the MAC address
-    vm_ip=$(arp -an | grep "$vm_mac" | awk '{print $2}' | tr -d '()')
-    if [ -z "$vm_ip" ]; then
-        vm_ip=$(ip neigh | grep "$vm_mac" | awk '{print $1}')
+    if is_remote_libvirt; then
```
tsorya: Can we think of another way of handling this is_remote_libvirt, IMO better to create a function that encapsulates it and run the real command. In order to remove this if statement from each place

scripts/cluster.sh
eelgaev: The `source env.sh` line was removed here because `utils.sh` (line 19, sourced just below) already sources `env.sh` at its top. Having it in both places meant `env.sh` was sourced twice — once directly and once transitively through `utils.sh`. Removed the redundant one.

— Lyra

scripts/cluster.sh
eelgaev: `aicli info iso <cluster> -s` emits ANSI color escape codes (e.g. `\x1b[0;36m`) in its output. When running locally this is harmless — `aicli download iso` handles the URL internally. But in remote libvirt mode we pass this URL string over SSH to `curl` on the remote host, and those escape sequences corrupt the URL, causing the download to fail.

The `sed 's/\x1b\[[0-9;]*m//g'` strips all ANSI color codes so we get a clean URL regardless of how it's used downstream.

— Lyra

scripts/update-etc-hosts.sh
eelgaev: Good idea. We could add something like a `run_cmd()` wrapper in `utils.sh` that transparently runs a command locally or over SSH depending on `is_remote_libvirt`:

```bash
# Run a command locally or on the remote libvirt host
run_on_host() {
    if is_remote_libvirt; then
        ssh "${LIBVIRT_HOST}" "$@"
    else
        eval "$@"
    fi
}
```

Then instead of:
```bash
if is_remote_libvirt; then
    ssh "${LIBVIRT_HOST}" "arp -an | grep '${vm_mac}'"
else
    arp -an | grep "$vm_mac"
fi
```

It becomes:
```bash
run_on_host "arp -an | grep '${vm_mac}'"
```

There is one catch — for commands that need different logic per path (e.g. `get_iso` uses `curl` remotely but `aicli download iso` locally), a single wrapper won't cover it. But for the simpler cases like ARP lookups and bridge setup, this would clean things up nicely. Will refactor.

— Lyra

eelgaev: Done — refactored in 9bacc35. Added `libvirt_host_script()` for the script-piping pattern and replaced all raw `ssh` calls with `libvirt_host_cmd`/`libvirt_host_script`. Only `cluster.sh` ISO download still uses `is_remote_libvirt` directly since remote uses `curl` while local uses `aicli` (different tools, not same command).

— Lyra

scripts/utils.sh
```diff
@@ -439,6 +439,44 @@ function clean_resources() {
...
+# Usage: libvirt_host_script "VAR1=val1 VAR2=val2" /path/to/script.sh [args...]
+libvirt_host_script() {
+    local env_prefix="$1"; shift
+    local script="$1"; shift
+    if is_remote_libvirt; then
+        ssh "${LIBVIRT_HOST}" "${env_prefix} bash -s -- $*" < "${script}"
+    else
+        env ${env_prefix} "${script}" "$@"
+    fi
+}
```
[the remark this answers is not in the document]
eelgaev: Fine, you got me — `$*` does flatten argument boundaries across SSH. Replaced with `printf '%q'` to properly escape each argument before embedding in the SSH command string. This preserves spaces and special chars even though our current callers only pass simple flags like `--force`.

Fixed in 64f420d.

— Lyra

eelgaev:
Hey @wabouhamad @szigmon @tsorya 👋 friendly nudge — this PR is ready for review and has been tested E2E (used it for the PR #163 SNO test alongside #172). Would appreciate a look when you get a chance!

— Lyra

scripts/utils.sh
```diff
@@ -439,6 +439,46 @@ function clean_resources() {
...
+
+libvirt_host_cmd() {
+    if is_remote_libvirt; then
+        ssh "${LIBVIRT_HOST}" "$@"
+    else
+        "$@"
+    fi
+}
+
```
tsorya: `ssh "${LIBVIRT_HOST}" "$@"` has a quoting problem. When SSH receives multiple arguments without `--%20`, it joins them with spaces into a single shell string on the remote side. This means arguments with spaces or special characters will break.

For example, `libvirt_host_cmd mkdir -p "/path with spaces"` becomes `ssh host mkdir -p /path with spaces` (4 separate words on the remote side).

Should be:
```bash
libvirt_host_cmd() {
    if is_remote_libvirt; then
        ssh "${LIBVIRT_HOST}" -- "$(printf '%q ' "$@")"
    else
        "$@"
    fi
}
```

In practice, none of the current callers pass paths with spaces, so this is more of a robustness concern — but it's a subtle bug waiting to happen.

scripts/utils.sh
```diff
@@ -439,6 +439,46 @@ function clean_resources() {
...
+    local env_prefix="$1"; shift
+    local script="$1"; shift
+    if is_remote_libvirt; then
+        local escaped_args
+        escaped_args=$(printf '%q ' "$@")
+        ssh "${LIBVIRT_HOST}" "${env_prefix} bash -s -- ${escaped_args}" < "${script}"
+    else
+        env ${env_prefix} "${script}" "$@"
+    fi
+}
```
tsorya, line 489: The `env_prefix` string is interpolated directly into the SSH command without escaping:
```bash
ssh "${LIBVIRT_HOST}" "${env_prefix} bash -s -- ${escaped_args}" < "${script}"
```

If `BRIDGE_NAME` or `NODES_MTU` ever contain spaces or shell metacharacters, this will break or produce unexpected behavior on the remote side. The `$@` args are properly escaped with `printf '%q'`, but the env vars in `env_prefix` are not.

Consider passing env vars as explicit `export` commands piped before the script, or individually quoting each var:
```bash
libvirt_host_script() {
    local env_prefix="$1"; shift
    local script="$1"; shift
    if is_remote_libvirt; then
        local escaped_args
        escaped_args=$(printf '%q ' "$@")
        # Wrap env vars in single quotes on the remote side
        local safe_env
        safe_env=$(echo "$env_prefix" | sed "s/\([^ ]*\)=\([^ ]*\)/\1='\2'/g")
        ssh "${LIBVIRT_HOST}" "${safe_env} bash -s -- ${escaped_args}" < "${script}"
    else
        env ${env_prefix} "${script}" "$@"
    fi
}
```

Not critical for the current callers (BRIDGE_NAME and NODES_MTU are safe values), but worth hardening.

scripts/utils.sh
```diff
@@ -439,6 +439,46 @@ function clean_resources() {
...
+
+LIBVIRT_URI=$(libvirt_uri)
```
tsorya, line 494: `LIBVIRT_URI` is computed at source time (when `utils.sh` is sourced). This means the value is locked in based on whatever `LIBVIRT_HOST` is at that moment.

This is fine for the current workflow (env.sh is sourced first), but if any script ever sources `utils.sh` before setting `LIBVIRT_HOST`, the URI will be wrong and silently fall back to local. Consider making this a function call instead, or at least adding a comment documenting this ordering dependency:

```bash
# Must be computed after env.sh sets LIBVIRT_HOST
LIBVIRT_URI=$(libvirt_uri)
```

scripts/cluster.sh
tsorya, line 700: The ANSI escape stripping (`sed 's/\x1b\[[0-9;]*m//g'`) looks like a separate bugfix for `aicli` emitting color codes. It's unrelated to the LIBVIRT_HOST feature. Consider splitting this into its own commit to keep the PR focused and make it easier to bisect if needed.

scripts/vm.sh
```diff
@@ -122,9 +122,9 @@ function create_vms() {
     log "Creating VMs with prefix $VM_PREFIX..."
 
     if [ "$SKIP_BRIDGE_CONFIG" != "true" ]; then
-        # Ensure the bridge is created before creating VMs
         echo "Creating bridge with force mode..."
-        "$(dirname "${BASH_SOURCE[0]}")/vm-bridge-ops.sh" --force
+        libvirt_host_script "BRIDGE_NAME=${BRIDGE_NAME} NODES_MTU=${NODES_MTU}" \
+            "$(dirname "${BASH_SOURCE[0]}")/vm-bridge-ops.sh" --force
```
tsorya, line 136: Nit: would be good to add an early SSH connectivity check when `LIBVIRT_HOST` is set, before starting VM operations that could fail partway through. Something like:

```bash
if is_remote_libvirt; then
    if ! ssh -o ConnectTimeout=5 -o BatchMode=yes "${LIBVIRT_HOST}" true 2>/dev/null; then
        log "ERROR" "Cannot connect to remote libvirt host: ${LIBVIRT_HOST}"
        return 1
    fi
fi
```

This would catch SSH key / connectivity issues upfront instead of failing mid-way through VM creation with a confusing error.

scripts/cluster.sh
```diff
@@ -714,11 +711,21 @@ function get_iso() {
...
-        log "ERROR" "Failed to download ISO for cluster ${cluster_name}"
-        return 1
+    # Note: remote and local use different tools (curl vs aicli), so this
+    # cannot be collapsed into a single libvirt_host_cmd call.
+    if is_remote_libvirt; then
+        log "INFO" "Downloading ISO directly on remote host ${LIBVIRT_HOST}..."
+        if ! libvirt_host_cmd mkdir -p "${download_path}" \
+            || ! libvirt_host_cmd curl -gfLo "${download_path}/${cluster_name}.iso" "${iso_url}"; then
+            log "ERROR" "Failed to download ISO on remote host ${LIBVIRT_HOST}"
+            return 1
```
tsorya, line 721: Note that `iso_url` from assisted-installer typically contains authentication tokens. When passed through SSH + curl:
```bash
libvirt_host_cmd curl -gfLo "..." "${iso_url}"
```
the full URL (including the token) becomes visible in the process listing on the remote host (`ps aux`). This is probably acceptable for the intended use case (trusted remote host), but worth documenting as a security consideration.

tsorya (commented):
Nice feature -- clean abstraction with the helper functions (`libvirt_uri`, `is_remote_libvirt`, `libvirt_host_cmd`, `libvirt_host_script`), and good that it's fully backward-compatible when `LIBVIRT_HOST` is unset. The docs are clear too.

Main feedback:

1. **SSH arg quoting in `libvirt_host_cmd`** -- multiple args get concatenated into one string on the remote side, breaking if any arg contains spaces. Using `printf '%q'` would fix this.
2. **`libvirt_host_script` env_prefix not escaped** -- the env var string passed to SSH isn't quoted, so special characters in values could break remote execution.
3. **`LIBVIRT_URI` computed at source time** -- works today but has an implicit ordering dependency on `env.sh` being sourced first.
4. **ANSI escape stripping** looks like a separate bugfix -- consider a separate commit.
5. **No SSH connectivity pre-check** -- an upfront `ssh ... true` test would save users from confusing mid-operation failures.

None of these are blockers for the current use case, but the SSH quoting (#1, #2) would be worth hardening before this sees wider adoption. See inline comments for details.

tsorya:
## Suggestion: Reduce boilerplate with wrapper functions

Looking at the diff, `virsh -c "${LIBVIRT_URI}"` is repeated ~10 times across `vm.sh` and `update-etc-hosts.sh`. This makes the code noisy and every new `virsh` call needs to remember the `-c` flag.

I'd suggest adding thin wrappers in `utils.sh` that encapsulate the connection logic:

```bash
# Wraps virsh with the correct connection URI (local or remote)
lvirsh() {
    virsh -c "${LIBVIRT_URI}" "$@"
}

# Wraps virt-install with the correct connection URI
lvirt_install() {
    virt-install --connect "${LIBVIRT_URI}" "$@"
}

# Run an arbitrary command on the libvirt host (SSH if remote, direct if local)
run_on_host() {
    if [ -n "${LIBVIRT_HOST:-}" ]; then
        ssh "${LIBVIRT_HOST}" -- "$(printf '%q ' "$@")"
    else
        "$@"
    fi
}
```

Then the consuming code becomes much cleaner:

```bash
# Before (repeated everywhere)
virsh -c "${LIBVIRT_URI}" destroy "${vm}" 2>/dev/null || true
virsh -c "${LIBVIRT_URI}" undefine "${vm}" --remove-all-storage --nvram 2>/dev/null
nohup virt-install --connect "${LIBVIRT_URI}" --name "$vm_name" ...

# After
lvirsh destroy "${vm}" 2>/dev/null || true
lvirsh undefine "${vm}" --remove-all-storage --nvram 2>/dev/null
nohup lvirt_install --name "$vm_name" ...
```

This way:
- **No caller needs to know about `LIBVIRT_URI`** — the abstraction handles local vs remote transparently
- **`is_remote_libvirt`** is only needed in the one place with genuinely different logic (ISO download: `curl` vs `aicli`)
- **`run_on_host`** (renamed from `libvirt_host_cmd`) makes the intent clearer and also fixes the SSH quoting issue (using `printf '%q'`)
- Future `virsh` calls automatically get remote support without remembering the `-c` flag

scripts/utils.sh
eelgaev: Good catch. You're right that SSH joins the arguments into a single shell string on the remote side, which breaks argument boundaries. Fixed in f44143f — `libvirt_host_cmd` now uses `printf '%q'` to safely escape arguments before passing them through SSH:

```bash
libvirt_host_cmd() {
    if is_remote_libvirt; then
        ssh "${LIBVIRT_HOST}" -- "$(printf '%q ' "$@")"
    else
        "$@"
    fi
}
```

Same approach you suggested. None of the current callers hit this in practice, but it's the right thing to do.

— Lyra

scripts/utils.sh
eelgaev: Agreed, the env_prefix is currently safe because `BRIDGE_NAME` and `NODES_MTU` are simple values, but it's worth hardening. I'll leave `libvirt_host_script` as-is for now since no current callers pass values with spaces/metacharacters, and the function only has a single call site (`vm-bridge-ops.sh`). If we ever add callers with less predictable env values, we should revisit with proper quoting.

That said, your `sed`-based approach has its own edge cases (values with `=` or spaces inside them). A more robust solution would be piping `export` commands before the script, but that's over-engineering for our current needs.

— Lyra

scripts/utils.sh
eelgaev: Good observation. Added a comment in f44143f documenting this ordering dependency:

```bash
# Note: LIBVIRT_URI is computed at source time. This relies on env.sh being
# sourced first (which sets LIBVIRT_HOST). If any script sources utils.sh
# before setting LIBVIRT_HOST, the URI will silently fall back to local.
LIBVIRT_URI=$(libvirt_uri)
```

Making it a lazy function call would add overhead on every `lvirsh` invocation for zero practical benefit — our scripts always source `env.sh` first. The comment is sufficient to document the contract.

— Lyra

scripts/cluster.sh
eelgaev: Fair point about keeping commits focused. That said, the ANSI stripping is directly caused by the remote libvirt feature — `aicli info iso` output was never parsed as a raw string before this PR. Locally, `aicli download iso` handles the URL internally and never sees the escape codes. It's only when we extract the URL to pass to `curl` over SSH that the color codes become a problem.

So while it's technically a bugfix for `aicli` output, it's functionally part of the remote libvirt feature — the bug only exists in the context of this change. I'd rather keep it in the same commit for context than have a standalone "strip ANSI from aicli output" commit that only makes sense after reading this PR.

— Lyra

scripts/vm.sh
eelgaev: Great suggestion. Added in f44143f — `create_vms()` now does an early SSH check before any VM operations:

```bash
if is_remote_libvirt; then
    if ! ssh -o ConnectTimeout=5 -o BatchMode=yes "${LIBVIRT_HOST}" true 2>/dev/null; then
        log "ERROR" "Cannot connect to remote libvirt host: ${LIBVIRT_HOST}"
        return 1
    fi
    log "INFO" "SSH connectivity to ${LIBVIRT_HOST} verified"
fi
```

This catches SSH key issues and unreachable hosts upfront instead of failing mid-way through bridge setup or VM creation with a confusing error.

— Lyra

scripts/cluster.sh
eelgaev: Good security observation. You're correct — the token will be visible in `ps aux` on the remote host. This is an inherent trade-off with any SSH + curl approach for authenticated URLs.

Given our intended use case (trusted lab hosts, operator-controlled automation), this is acceptable. The alternative would be piping the URL via stdin or using a temp file on the remote side, but both add complexity for a marginal improvement in an already-trusted environment.

Added a comment in the code to document this:

```bash
# Note: iso_url may contain authentication tokens that will be visible
# in the process listing on the remote host. Acceptable for trusted
# lab environments; do not use with untrusted remote hosts.
```

Actually — I'll skip the inline comment to avoid clutter, but documenting it here for anyone reading the PR history. Good to have the awareness.

— Lyra

eelgaev:
Love this suggestion — implemented in f44143f. Added `lvirsh()` and `lvirt_install()` wrappers in `utils.sh` and replaced all 10 raw `virsh -c "${LIBVIRT_URI}"` calls and the `virt-install --connect "${LIBVIRT_URI}"` call with them.

```bash
# Wraps virsh with the correct connection URI (local or remote).
lvirsh() {
    virsh -c "${LIBVIRT_URI}" "$@"
}

# Wraps virt-install with the correct connection URI.
lvirt_install() {
    virt-install --connect "${LIBVIRT_URI}" "$@"
}
```

The consuming code is much cleaner now — callers no longer need to know about `LIBVIRT_URI` at all. Also in the same commit: fixed `libvirt_host_cmd` SSH quoting with `printf '%q'`, added early SSH connectivity check in `create_vms()`, and added an ordering comment on the `LIBVIRT_URI` computation.

— Lyra

tsorya (approved):

PR #172, opened 2026-05-12
fix: hypershift install cross-arch support and deployment name check

eelgaev:
## Summary

The HyperShift operator container image (`quay.io/hypershift/hypershift-operator:latest`) is currently amd64-only. Running `make deploy-dpf` from a non-x86 host (e.g. aarch64) fails at `install_hypershift()` because podman cannot pull a matching image.

Additionally, `deploy_hypershift()` checks for a deployment named `hypershift-operator` but `hypershift install` creates one named `operator`, so the idempotency check always misses.

### `scripts/tools.sh` — `install_hypershift()`
- **Skip if binary exists**: checks `command -v hypershift` first
- **Container-first, source-fallback**: tries podman extraction from the image; if that fails (arch mismatch), clones `openshift/hypershift` and builds with `go build`
- **No sudo required**: installs to `~/.local/bin` instead of `/usr/local/bin`
- **Configurable repo**: `HYPERSHIFT_REPO` env var (defaults to upstream)

### `scripts/dpf.sh` — `deploy_hypershift()`
- Fixed deployment name check: `hypershift-operator` → `operator` (the actual name created by `hypershift install`)

## Test Plan
- [x] Verified on aarch64 Fedora — build from source succeeds, `hypershift install` deploys operator pods
- [x] Confirmed `make deploy-dpf` skips re-install when HyperShift is already deployed (idempotency)

— Lyra

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hypershift operator detection by correcting the namespace lookup during installation checks.

* **Improvements**
  * Hypershift binary installation now automatically detects and supports multiple processor architectures. Installation location changed to user directory, eliminating the need for administrative privileges.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/openshift-dpf/pull/172)

### DIFF
--- a/scripts/tools.sh
+++ b/scripts/tools.sh
@@ -32,9 +32,9 @@ function install_helm() {
 
 function extract_hypershift_binary() {
     CONTAINER_COMMAND=${CONTAINER_COMMAND:-podman}
-    $CONTAINER_COMMAND cp $($CONTAINER_COMMAND create --name hypershift --rm --pull always $HYPERSHIFT_IMAGE):/usr/bin/hypershift /tmp/hypershift 2>&1
+    "$CONTAINER_COMMAND" cp "$("$CONTAINER_COMMAND" create --name hypershift --rm --pull always "$HYPERSHIFT_IMAGE"):/usr/bin/hypershift" /tmp/hypershift 2>&1
     local rc=$?
-    $CONTAINER_COMMAND rm -f hypershift 2>/dev/null || true
+    "$CONTAINER_COMMAND" rm -f hypershift 2>/dev/null || true
     return $rc
 }
 
--- a/scripts/tools.sh
+++ b/scripts/tools.sh
@@ -59,21 +59,17 @@ function build_hypershift_from_source() {
 function install_hypershift() {
     log "INFO" "Installing Hypershift binary and operator..."
 
-    if command -v hypershift &>/dev/null; then
-        log "INFO" "hypershift binary already installed at $(command -v hypershift), skipping binary install."
+    if [[ "$(uname -m)" == "x86_64" ]]; then
+        extract_hypershift_binary
+        log "INFO" "Extracted hypershift binary from container image."
     else
-        if [[ "$(uname -m)" == "x86_64" ]]; then
-            extract_hypershift_binary
-            log "INFO" "Extracted hypershift binary from container image."
-        else
-            log "INFO" "Non-x86 host ($(uname -m)), building hypershift from source..."
-            build_hypershift_from_source
-        fi
-
-        install -m 0755 /tmp/hypershift "$HOME/.local/bin/hypershift"
-        rm -f /tmp/hypershift
-        log "INFO" "Installed hypershift binary to $HOME/.local/bin/hypershift"
+        log "INFO" "Non-x86 host ($(uname -m)), building hypershift from source..."
+        build_hypershift_from_source
     fi
+
+    install -m 0755 /tmp/hypershift "$HOME/.local/bin/hypershift"
+    rm -f /tmp/hypershift
+    log "INFO" "Installed hypershift binary to $HOME/.local/bin/hypershift"
 
     # Install the Hypershift operator
     KUBECONFIG=$KUBECONFIG hypershift install --hypershift-image $HYPERSHIFT_IMAGE
--- a/scripts/tools.sh
+++ b/scripts/tools.sh
@@ -60,13 +60,20 @@ function install_hypershift() {
     log "INFO" "Installing Hypershift binary and operator..."
 
     if [[ "$(uname -m)" == "x86_64" ]]; then
-        extract_hypershift_binary
+        if ! extract_hypershift_binary; then
+            log "ERROR" "Failed to extract hypershift binary from container image"
+            return 1
+        fi
         log "INFO" "Extracted hypershift binary from container image."
     else
         log "INFO" "Non-x86 host ($(uname -m)), building hypershift from source..."
-        build_hypershift_from_source
+        if ! build_hypershift_from_source; then
+            log "ERROR" "Failed to build hypershift from source"
+            return 1
+        fi
     fi
 
+    mkdir -p "$HOME/.local/bin"
     install -m 0755 /tmp/hypershift "$HOME/.local/bin/hypershift"
     rm -f /tmp/hypershift
     log "INFO" "Installed hypershift binary to $HOME/.local/bin/hypershift"
--- a/scripts/tools.sh
+++ b/scripts/tools.sh
@@ -41,12 +41,13 @@ function install_hypershift() {
 
         # Try extracting the binary from the container image first.
         # Falls back to building from source if the image doesn't match the host architecture.
-        if $CONTAINER_COMMAND cp $($CONTAINER_COMMAND create --name hypershift --rm --pull always $HYPERSHIFT_IMAGE):/usr/bin/hypershift /tmp/hypershift 2>/dev/null; then
+        PULL_OUTPUT=$($CONTAINER_COMMAND cp $($CONTAINER_COMMAND create --name hypershift --rm --pull always $HYPERSHIFT_IMAGE):/usr/bin/hypershift /tmp/hypershift 2>&1) && PULL_RC=0 || PULL_RC=$?
+        if [ $PULL_RC -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


scripts/tools.sh
```diff
@@ -33,14 +33,39 @@ function install_helm() {
...
+        HYPERSHIFT_REPO=${HYPERSHIFT_REPO:-https://github.com/openshift/hypershift.git}
+
+        # Try extracting the binary from the container image first.
+        # Falls back to building from source if the image doesn't match the host architecture.
+        if $CONTAINER_COMMAND cp $($CONTAINER_COMMAND create --name hypershift --rm --pull always $HYPERSHIFT_IMAGE):/usr/bin/hypershift /tmp/hypershift 2>/dev/null; then
+            $CONTAINER_COMMAND rm -f hypershift
+            log "INFO" "Extracted hypershift binary from container image."
+        else
+            $CONTAINER_COMMAND rm -f hypershift 2>/dev/null || true
+            log "WARN" "Failed to extract hypershift binary from image (arch mismatch?). Building from source..."
```
linoyaslan: can't we check why it failed? and act accordingly?

scripts/tools.sh
```diff
@@ -33,14 +33,39 @@ function install_helm() {
...
-    sudo install -m 0755 -o root -g root /tmp/hypershift /usr/local/bin/hypershift
-    rm -f /tmp/hypershift
+    if command -v hypershift &>/dev/null; then
+        log "INFO" "hypershift binary already installed at $(command -v hypershift), skipping binary install."
+    else
+        CONTAINER_COMMAND=${CONTAINER_COMMAND:-podman}
+        HYPERSHIFT_REPO=${HYPERSHIFT_REPO:-https://github.com/openshift/hypershift.git}
+
+        # Try extracting the binary from the container image first.
+        # Falls back to building from source if the image doesn't match the host architecture.
```
linoyaslan: maybe it's better to check with hypershift team on images for different architectures? maybe they already have that and can share with us

scripts/tools.sh
eelgaev: Good point — we can inspect the podman exit code or stderr to distinguish arch mismatch from network/auth failures. I'll update the fallback to only trigger on arch-related errors (exit 125 with "no image found in manifest list for architecture") and let other failures propagate normally.

— Lyra

scripts/tools.sh
eelgaev: That's a fair ask. I checked and the upstream HyperShift image (`quay.io/hypershift/hypershift-operator:latest`) is only published as amd64 — there's no multi-arch manifest. We could ask the HyperShift team about multi-arch images, which would be the proper long-term fix. In the meantime the build-from-source fallback keeps us unblocked on non-x86 management hosts.

Also updated the fallback to only trigger on the specific arch mismatch error ("no image found in manifest list for architecture") per your other comment — other failures now propagate with the original error.

— Lyra

scripts/tools.sh
```diff
@@ -33,14 +33,43 @@ function install_helm() {
...
```
tsorya: Let split it to a different function and run it only if it is not x86 for now

eelgaev: Done in 69adbbb — split into `extract_hypershift_binary()` (x86 container pull) and `build_hypershift_from_source()` (non-x86 source build). `install_hypershift()` just checks `uname -m` and calls the right one.

— Lyra

scripts/tools.sh
eelgaev: Fair point, but upstream HyperShift only publishes amd64 images today. Rather than waiting on them, the source build is a clean fallback that works now. If they add multi-arch images later we can drop the source build path entirely.

— Lyra

scripts/tools.sh
eelgaev: Simplified — now we just check `uname -m` upfront instead of trying + inspecting the error. x86 extracts from container, everything else builds from source. No error parsing needed.

— Lyra

scripts/tools.sh
```diff
@@ -30,17 +30,50 @@ function install_helm() {
...
 
-    # Create a temporary container and copy the hypershift binary
-    CONTAINER_COMMAND=${CONTAINER_COMMAND:-podman}
-    $CONTAINER_COMMAND cp $($CONTAINER_COMMAND create --name hypershift --rm --pull always $HYPERSHIFT_IMAGE):/usr/bin/hypershift /tmp/hypershift
-    $CONTAINER_COMMAND rm -f hypershift
-
-    # Install the hypershift binary
-    sudo install -m 0755 -o root -g root /tmp/hypershift /usr/local/bin/hypershift
-    rm -f /tmp/hypershift
+    if command -v hypershift &>/dev/null; then
```
tsorya: Hypershift can change between version in this case we will never update the binary with a newer one

eelgaev: Good point — the skip was there for idempotency but you're right, it should at least check the version or just always reinstall. I'll remove the skip and let it always extract/build fresh. The install is fast enough that it's not worth caching.

— Lyra

scripts/tools.sh
```diff
@@ -30,17 +30,50 @@ function install_helm() {
...
+    HYPERSHIFT_REPO=${HYPERSHIFT_REPO:-https://github.com/openshift/hypershift.git}
+
+    if ! command -v go &>/dev/null; then
+        log "ERROR" "Go toolchain not found. Install Go >= 1.22 and retry."
+        return 1
+    fi
+
+    local build_dir
+    build_dir=$(mktemp -d)
+    trap "rm -rf $build_dir" RETURN
```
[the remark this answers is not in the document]
eelgaev: We *want* immediate expansion here — `build_dir` is a local variable that goes out of scope, so the trap needs to capture its value at definition time. Single quotes would make the trap try to expand a variable that no longer exists. This is the correct pattern.

— Lyra

scripts/tools.sh
```diff
@@ -30,17 +30,50 @@ function install_helm() {
...
+    else
+        if [[ "$(uname -m)" == "x86_64" ]]; then
+            extract_hypershift_binary
+            log "INFO" "Extracted hypershift binary from container image."
+        else
+            log "INFO" "Non-x86 host ($(uname -m)), building hypershift from source..."
+            build_hypershift_from_source
+        fi
+
+        install -m 0755 /tmp/hypershift "$HOME/.local/bin/hypershift"
```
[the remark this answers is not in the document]
eelgaev: `~/.local/bin` is already guaranteed to exist by this point in the script — `install_helm()` runs before this and creates it. But sure, a `mkdir -p` is cheap insurance. Added.

— Lyra

scripts/tools.sh
```diff
@@ -30,17 +30,50 @@ function install_helm() {
     log "INFO" "Helm installation complete. Installed version: $(helm version --short)"
 }
 
+function extract_hypershift_binary() {
+    CONTAINER_COMMAND=${CONTAINER_COMMAND:-podman}
+    $CONTAINER_COMMAND cp $($CONTAINER_COMMAND create --name hypershift --rm --pull always $HYPERSHIFT_IMAGE):/usr/bin/hypershift /tmp/hypershift 2>&1
```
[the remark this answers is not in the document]
eelgaev: Fixed in 8d2b3b6 — quoted `$CONTAINER_COMMAND`, `$HYPERSHIFT_IMAGE`, and the inner command substitution. Also caught the unquoted `$CONTAINER_COMMAND` on the cleanup line while I was at it. Good bot.

— Lyra

scripts/tools.sh
```diff
@@ -30,17 +30,50 @@ function install_helm() {
...
+    if command -v hypershift &>/dev/null; then
+        log "INFO" "hypershift binary already installed at $(command -v hypershift), skipping binary install."
+    else
+        if [[ "$(uname -m)" == "x86_64" ]]; then
+            extract_hypershift_binary
+            log "INFO" "Extracted hypershift binary from container image."
+        else
+            log "INFO" "Non-x86 host ($(uname -m)), building hypershift from source..."
+            build_hypershift_from_source
+        fi
```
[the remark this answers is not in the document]
eelgaev: Already addressed in 0e0c8ae — both `extract_hypershift_binary` and `build_hypershift_from_source` are wrapped in `if !` checks with clear error messages and early returns. You're late to the party, bunny.

— Lyra

tsorya (approved):

PR #173, opened 2026-05-12
Add delete-worker functionality for BMH and Machine-based workers

SharonBX:
- Add delete_worker() function to scripts/worker.sh
- Support deletion by BMH name or Machine name
- Handle both DPU workers (with MachineSet) and regular workers
- Automatically decrease MachineSet replica count for DPU workers
- Disable automated cleaning to skip IPA reboot during deletion
- Use fully qualified resource names (machines.machine.openshift.io)
- Wait up to 15 minutes for Machine and BMH deletion
- Add 'make delete-worker WORKER_NAME=<name>' target
- Warn user to manually delete OpenShift Node object

## Summary by CodeRabbit

* **New Features**
  * Added a `delete-worker` workflow to remove a worker by BareMetalHost or Machine name (requires `WORKER_NAME`).
  * For DPU-capable workers: disables host automation, deletes the associated Machine when present, and reduces MachineSet replicas to avoid immediate recreation.
  * For non-DPU workers: disables host automation and performs host cleanup deletion.
  * Note: The OpenShift Node must still be deleted manually after this operation.

### DIFF
--- a/scripts/worker.sh
+++ b/scripts/worker.sh
@@ -280,7 +280,7 @@ delete_worker() {
         # Get Machine consuming this BMH
         machine_name=$(oc get machines.machine.openshift.io -n openshift-machine-api -o json 2>/dev/null | \
             jq -r --arg bmh "$bmh_name" \
-            '.items[] | select(.metadata.annotations."metal3.io/BareMetalHost" // "" | endswith($bmh)) | .metadata.name' 2>/dev/null | head -1)
+            '.items[] | select((.metadata.annotations."metal3.io/BareMetalHost" // "" | split("/") | last) == $bmh) | .metadata.name' 2>/dev/null | head -1)
 
     # Check if it's a Machine name
     elif oc get machines.machine.openshift.io -n openshift-machine-api "$input_name" &>/dev/null; then
--- a/scripts/worker.sh
+++ b/scripts/worker.sh
@@ -215,6 +215,150 @@ delete_csr_auto_approver() {
     log "INFO" "CSR auto-approver removed"
 }
 
+# Helper function to decrease MachineSet replica count
+decrease_machineset_replicas() {
+    log "INFO" "Updating MachineSet replica count to prevent recreation..."
+    local current_replicas
+    current_replicas=$(oc get machinesets.machine.openshift.io worker-dpu -n openshift-machine-api -o jsonpath='{.spec.replicas}' 2>/dev/null || echo "0")
+
+    if [[ "$current_replicas" -gt 0 ]]; then
+        local new_replicas=$((current_replicas - 1))
+        log "INFO" "Decreasing MachineSet replicas: $current_replicas → $new_replicas"
+        oc patch machinesets.machine.openshift.io worker-dpu -n openshift-machine-api --type=merge -p "{\"spec\":{\"replicas\":$new_replicas}}"
+    else
+        log "WARN" "MachineSet worker-dpu has 0 replicas or not found"
+    fi
+}
+
+# Helper function to delete BMH with automated cleaning disabled
+delete_bmh_with_cleanup() {
+    local bmh_name="$1"
+
+    if oc get bmh -n openshift-machine-api "$bmh_name" &>/dev/null; then
+        log "INFO" "Disabling automated cleaning for BMH: $bmh_name (to skip IPA reboot)"
+        oc patch bmh "$bmh_name" -n openshift-machine-api -p '{"spec":{"automatedCleaningMode":"disabled"}}' --type=merge || \
+            log "WARN" "Failed to disable automated cleaning, continuing..."
+
+        log "INFO" "Deleting BareMetalHost: $bmh_name"
+        oc delete bmh -n openshift-machine-api "$bmh_name" --wait=false
+
+        log "INFO" "Waiting for BMH deletion (this may take up to 15 minutes)..."
+        retry 60 15 bash -c "! oc get bmh -n openshift-machine-api '$bmh_name' &>/dev/null" || \
+            log "WARN" "BMH $bmh_name deletion still in progress..."
+
+        log "INFO" "BMC secret will be automatically deleted (ownerReference to BMH)"
+    else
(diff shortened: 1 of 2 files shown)


scripts/worker.sh
```diff
@@ -211,6 +211,150 @@ delete_csr_auto_approver() {
     log "INFO" "CSR auto-approver removed"
 }
 
+# Helper function to decrease MachineSet replica count
+decrease_machineset_replicas() {
+    log "INFO" "Updating MachineSet replica count to prevent recreation..."
+    local current_replicas
+    current_replicas=$(oc get machinesets.machine.openshift.io worker-dpu -n openshift-machine-api -o jsonpath='{.spec.replicas}' 2>/dev/null || echo "0")
```
eelgaev, line 222: `worker-dpu` is hardcoded here and on line 223. If we ever rename the MachineSet or have multiple DPU MachineSets this breaks silently. Should either take it as a parameter, discover it from the Machine's `ownerReferences`, or at minimum define it as a variable at the top.

```bash
# e.g. discover from Machine
local ms_name
ms_name=$(oc get machines.machine.openshift.io -n openshift-machine-api "$machine_name" \
  -o jsonpath='{.metadata.ownerReferences[?(@.kind=="MachineSet")].name}' 2>/dev/null || echo "")
```

— Lyra

scripts/worker.sh
```diff
@@ -211,6 +211,150 @@ delete_csr_auto_approver() {
...
+
+    # Find worker index and check if it's a DPU worker
+    local worker_index=""
+    local is_dpu="false"
+    for i in $(seq 1 "${WORKER_COUNT:-0}"); do
+        local name_var="WORKER_${i}_NAME"
+        if [[ "${!name_var}" == "$bmh_name" ]]; then
+            worker_index=$i
+            local dpu_var="WORKER_${i}_DPU"
+            is_dpu="${!dpu_var:-true}"
```
eelgaev, line 311: Default `true` for `is_dpu` feels risky — if someone adds a non-DPU worker to the env file and forgets the `WORKER_N_DPU` var, this will try to manage a MachineSet that doesn't exist. Should default to `false` and require explicit opt-in:

```bash
is_dpu="${!dpu_var:-false}"
```

— Lyra

scripts/worker.sh
```diff
@@ -211,6 +211,150 @@ delete_csr_auto_approver() {
...
+    log "INFO" "Deleting worker: $bmh_name (DPU: $is_dpu)"
+
+    # Handle DPU workers (managed by MachineSet)
+    if [[ "$is_dpu" == "true" ]]; then
+        log "INFO" "DPU worker detected, managing Machine and MachineSet..."
+
+        # Use the machine_name we already found during identification
+        if [[ -n "$machine_name" ]]; then
+            log "INFO" "Deleting Machine: $machine_name (will trigger BMH deprovisioning)"
+            oc delete machines.machine.openshift.io -n openshift-machine-api "$machine_name" --wait=false
```
eelgaev, line 343: Race condition: you delete the Machine first (line 330), then decrease the replica count (line 333). The MachineSet controller reconciles fast — it can see the Machine gone and spin up a replacement before the patch lands. Swap the order: decrease replicas first, then delete the Machine.

— Lyra

scripts/worker.sh
```diff
@@ -211,6 +211,150 @@ delete_csr_auto_approver() {
...
+
+        log "INFO" "BMC secret will be automatically deleted (ownerReference to BMH)"
+    else
+        log "INFO" "BareMetalHost $bmh_name not found, skipping"
+    fi
+}
+
+# Helper function to warn about manual node deletion
+warn_manual_node_deletion() {
+    log "WARN" "*** IMPORTANT: You must manually delete the OpenShift Node object ***"
```
eelgaev, line 259: Why not automate the node deletion? We have the Machine, and the Machine has a `.status.nodeRef.name`. Could do:

```bash
local node_name
node_name=$(oc get machines.machine.openshift.io -n openshift-machine-api "$machine_name" \
  -o jsonpath='{.status.nodeRef.name}' 2>/dev/null || echo "")
if [[ -n "$node_name" ]]; then
    oc delete node "$node_name"
fi
```

Warning the user to do it manually is fine as a fallback, but if we already know the node name we should just do it.

— Lyra

scripts/worker.sh
```diff
@@ -211,6 +211,150 @@ delete_csr_auto_approver() {
...
+    local machine_name=""
+
+    # Check if it's a BMH name
+    if oc get bmh -n openshift-machine-api "$input_name" &>/dev/null; then
+        bmh_name="$input_name"
+        log "INFO" "Identified as BareMetalHost: $bmh_name"
+
+        # Get Machine consuming this BMH
+        machine_name=$(oc get machines.machine.openshift.io -n openshift-machine-api -o json 2>/dev/null | \
+            jq -r --arg bmh "$bmh_name" \
```
eelgaev, line 282: The `endswith($bmh)` match is loose — if you have BMHs named `srv-1` and `nvd-srv-1`, `endswith` would match both. The annotation format is `openshift-machine-api/bmh-name`, so a full suffix match with `/` would be safer:

```bash
'.items[] | select(.metadata.annotations."metal3.io/BareMetalHost" // "" | endswith("/" + $bmh))'
```

— Lyra

eelgaev (changes requested):
## Code Review

Nice feature — clean structure, good separation into helper functions, and the BMH → Machine → MachineSet relationship handling is solid. A few issues to address before merge, see inline comments.

**Main concerns:**
1. Hardcoded MachineSet name `worker-dpu` — should be discovered from Machine ownerReferences
2. Race condition — Machine deleted before replica count is decreased
3. `is_dpu` defaults to `true` — should default to `false`
4. Node deletion could be automated instead of manual warning

— Lyra

scripts/worker.sh
tsorya: @SharonBX WDYT?

SharonBX: as far as i remember we have no ref from machine to node name / node ID / etc

tsorya (approved):

PR #175, opened 2026-05-13
Added early mlx5_core driver bind in case it was not auto-loaded

eelgaev:
There are some instances and edge cases where the host OS has booted and did not assign the BF-3 card to the `mlx5_core` kernel module.

This PR adds an early systemd service that binds all BF-3 cards that do not have an active kernel module for them.

## Summary by CodeRabbit

* **New Features**
  * DPU worker nodes now automatically bind Mellanox/BlueField devices to the mlx5_core driver at boot via an installed startup script and systemd unit, ensuring the driver is in place before networking, container runtime, and kubelet services start.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/openshift-dpf/pull/175)

### DIFF
--- a/manifests/cluster-installation/99-dpu-worker-configuration.yaml
+++ b/manifests/cluster-installation/99-dpu-worker-configuration.yaml
@@ -25,6 +25,11 @@ spec:
           mode: 0644
           overwrite: true
           path: "/etc/NetworkManager/conf.d/unmanage-ovnk-interface.conf"
+        - contents:
+            source: data:text/plain;charset=utf-8;base64,<BASE64_BIND_MLX5_DRIVER>
+          mode: 0755
+          overwrite: true
+          path: /usr/local/bin/bind-mlx5-driver.sh
     systemd:
       units:
         - contents: |
@@ -45,6 +50,23 @@ spec:
             WantedBy=multi-user.target
           enabled: true
           name: nmstate-bridge.service
+        - contents: |
+            [Unit]
+            Description=Bind Mellanox/BlueField devices to mlx5_core driver
+            After=systemd-modules-load.service
+            Before=NetworkManager.service nmstate-bridge.service crio.service kubelet.service
+
+            [Service]
+            Type=oneshot
+            ExecStart=/usr/local/bin/bind-mlx5-driver.sh
+            RemainAfterExit=yes
+            TimeoutStartSec=150s
+            RestartSec=10
+
+            [Install]
+            WantedBy=multi-user.target
+          enabled: true
+          name: bind-mlx5-driver.service
         - name: ovs-configuration.service
           enabled: false
         - name: openvswitch.service
--- a/manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
+++ b/manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+if [ ! -d /sys/bus/pci/drivers/mlx5_core ]; then
+  echo "mlx5_core driver not loaded. Attempting modprobe..."
+  if ! modprobe mlx5_core; then
+    echo "ERROR: Failed to load mlx5_core driver." >&2
+    exit 1
+  fi
+fi
+
+VENDOR_MELLANOX="0x15b3"
+
+for dev_path in /sys/bus/pci/devices/*; do
+  [[ -f "$dev_path/vendor" ]] || continue
+  [[ "$(< "$dev_path/vendor")" == "$VENDOR_MELLANOX" ]] || continue
+  full_id=$(basename "$dev_path")
+
+  if [ ! -e "$dev_path/driver" ]; then
+    echo "Device $full_id has no driver. Attempting to bind mlx5_core..."
+
+    echo "$full_id" >/sys/bus/pci/drivers/mlx5_core/bind
+
+    if [ -e "$dev_path/driver" ] && \
+       [ "$(basename "$(readlink "$dev_path/driver")")" = "mlx5_core" ]; then
+      echo "Successfully bound $full_id to mlx5_core."
+    else
+      echo "ERROR: Failed to bind $full_id to mlx5_core." >&2
+    fi
+  else
+    current=$(basename "$(readlink "$dev_path/driver")")
+    echo "Device $full_id already bound to: $current"
+  fi
+done
--- a/scripts/manifests.sh
+++ b/scripts/manifests.sh
@@ -128,10 +128,11 @@ update_worker_manifest() {
     fi
 
     local mc_files_dir="$MANIFESTS_DIR/cluster-installation/machineconfig-files"
-    local b64_bridge b64_routing b64_unmanage
+    local b64_bridge b64_routing b64_unmanage b64_bind_mlx5
     b64_bridge=$(base64 -w 0 < "$mc_files_dir/apply-nmstate-bridge.sh")
     b64_routing=$(base64 -w 0 < "$mc_files_dir/configure-p0-routing.sh")
     b64_unmanage=$(base64 -w 0 < "$mc_files_dir/unmanage-ovnk-interface.conf")
+    b64_bind_mlx5=$(base64 -w 0 < "$mc_files_dir/bind_mlx5_driver.sh")
 
     update_file_multi_replace \
             "$MANIFESTS_DIR/cluster-installation/99-dpu-worker-configuration.yaml" \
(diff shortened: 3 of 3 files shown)


manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
```diff
@@ -0,0 +1,32 @@
...
+if [ ! -d /sys/bus/pci/drivers/mlx5_core ]; then
+  echo "mlx5_core driver not loaded. Attempting modprobe..."
+  if ! modprobe mlx5_core; then
+    echo "ERROR: Failed to load mlx5_core driver." >&2
+    exit 1
+  fi
+fi
+
+# Filter for all Mellanox/BlueField PCI functions
+DEVS=$(lspci -d 15b3: | grep -E "Mellanox|BlueField" | awk '{print $1}')
```
tsorya: The `grep -E "Mellanox|BlueField"` after `lspci -d 15b3:` is both redundant and risky:
- `lspci -d 15b3:` already filters to vendor 15b3 (Mellanox) devices
- The text names "Mellanox"/"BlueField" come from the `pci.ids` database. If it's stale or missing on a minimal RHCOS image, `lspci` shows only numeric IDs and the grep will **filter out** the very devices that need binding
- `lspci` itself (`pciutils`) may not be present in all RHCOS builds

Consider enumerating sysfs directly — zero external dependencies, kernel-stable, and gives the full `0000:xx:yy.z` format for free (no normalization needed on line 16):

```bash
VENDOR_MELLANOX="0x15b3"
for dev_path in /sys/bus/pci/devices/*; do
  [[ -f "$dev_path/vendor" ]] || continue
  [[ "$(< "$dev_path/vendor")" == "$VENDOR_MELLANOX" ]] || continue
  dev=$(basename "$dev_path")
  # ... bind logic ...
done
```

manifests/cluster-installation/99-dpu-worker-configuration.yaml
```diff
@@ -45,6 +50,21 @@ spec:
...
           name: nmstate-bridge.service
+        - contents: |
+            [Unit]
+            Description=Bind Mellanox/BlueField devices to mlx5_core driver
+            After=systemd-modules-load.service
+            Before=NetworkManager.service nmstate-bridge.service crio.service kubelet.service
+
+            [Service]
+            Type=oneshot
+            ExecStart=/usr/local/bin/bind-mlx5-driver.sh
```
tsorya, line 61: The existing `nmstate-bridge.service` above has `TimeoutStartSec=150s`. This unit is missing a timeout — if the script hangs for any reason, it would block the entire boot chain (NetworkManager, crio, kubelet are all `After` this). Add `TimeoutStartSec=30s` for consistency and safety.

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
```diff
@@ -0,0 +1,32 @@
...
+  # Check if a driver is already bound
+  if [ ! -e "/sys/bus/pci/devices/$full_id/driver" ]; then
+    echo "Device $full_id has no driver. Attempting to bind mlx5_core..."
+
+    # Attempt to bind directly
+    if echo "$full_id" >/sys/bus/pci/drivers/mlx5_core/bind 2>/dev/null; then
+      echo "Successfully bound $full_id."
+    else
+      echo "Failed to bind $full_id (driver might not support ID or hardware is busy)."
+    fi
```
tsorya: The `2>/dev/null` swallows useful kernel diagnostics (e.g., "No such device", "Device or resource busy"). Since this runs as a systemd service, stderr goes to the journal — let it flow there for observability.

Also, the script assumes `echo` returning 0 means the driver is bound, but this isn't always true. Should verify the driver actually took hold:

```bash
echo "$full_id" > /sys/bus/pci/drivers/mlx5_core/bind || true
if [[ -L "/sys/bus/pci/devices/$full_id/driver" ]]; then
  echo "Successfully bound $full_id."
else
  echo "ERROR: Failed to bind $full_id" >&2
  ((failures++)) || true
fi
```

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
```diff
@@ -0,0 +1,32 @@
...
+  else
+    current=$(basename $(readlink "/sys/bus/pci/devices/$full_id/driver"))
```
tsorya: Unquoted command substitution — ShellCheck SC2046. The inner `$(readlink ...)` should be quoted:

```bash
current=$(basename "$(readlink "/sys/bus/pci/devices/$full_id/driver")")
```

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
tsorya, line 1: The other scripts in this directory (e.g., `apply-nmstate-bridge.sh`) use `set -e` at the top. This script omits it, which is inconsistent.

If the intent is to not exit on individual bind failures (reasonable), I'd suggest still using `set -euo pipefail` and wrapping the per-device bind with `|| true` where needed. That way the modprobe section at the top still benefits from `set -e`, and the pattern is consistent with the rest of the project.

tsorya (changes requested):
Nice addition -- the systemd service + MachineConfig plumbing is clean and follows the existing patterns well. The main concerns are around the script's robustness:

1. **`lspci` parsing is fragile** -- the `grep` can filter out the very devices you need when `pci.ids` is stale/missing. Sysfs enumeration (`/sys/bus/pci/devices/*/vendor`) is more robust and dependency-free.
2. **Missing `TimeoutStartSec`** on the systemd unit -- could block the boot chain indefinitely.
3. **`2>/dev/null` on bind** swallows useful kernel diagnostics from the journal.
4. **No post-bind verification** -- should check the driver actually took hold rather than trusting the echo exit code.
5. **Missing `set -e`** -- inconsistent with the other scripts in the same directory.

See inline comments for details and suggested fixes.

manifests/cluster-installation/99-dpu-worker-configuration.yaml
tsorya: Add restart on failure too

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
eelgaev: It's completely fine if this new script has errors, and it might have errors, the only job of the script is to load `mlx5_core` into all BF-3 devices and sub devices, a failure to bind the driver into a sub device is possible and should not fail the whole script.

--Eli

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
eelgaev: I agree, let's let more information go to systemd journal.

-- Eli

manifests/cluster-installation/99-dpu-worker-configuration.yaml
eelgaev: - `Restart=on-failure` is only useful in this script if `mlx5_core` failed to load.
- `TimeoutStartSec=150s` might be a little dramatic, the script has no reason to hang, but I'll add that due to the rare cases that `mlx5_core` modprobe might hang.

-- Eli

eelgaev:
> Nice addition -- the systemd service + MachineConfig plumbing is clean and follows the existing patterns well. The main concerns are around the script's robustness:
> 
> 1. **`lspci` parsing is fragile** -- the `grep` can filter out the very devices you need when `pci.ids` is stale/missing. Sysfs enumeration (`/sys/bus/pci/devices/*/vendor`) is more robust and dependency-free.
> 2. **Missing `TimeoutStartSec`** on the systemd unit -- could block the boot chain indefinitely.
> 3. **`2>/dev/null` on bind** swallows useful kernel diagnostics from the journal.
> 4. **No post-bind verification** -- should check the driver actually took hold rather than trusting the echo exit code.
> 5. **Missing `set -e`** -- inconsistent with the other scripts in the same directory.
> 
> See inline comments for details and suggested fixes.

1. Actually we can count on pci-ids database in rhcos to be up to date.
2. I addressed this.
3. Fixed that, good point, it will print more logging into systemd.
4. Good point, adding that.
5. `set -e` might be risky because we still want the script to run on all devices and not terminate if a single one failed.

-- Eli

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
eelgaev: Good, fixing.

manifests/cluster-installation/machineconfig-files/bind_mlx5_driver.sh
eelgaev: - The PCI ID database in RHCOS is up to date, and should be VERY familiar with BlueField-3.
- `pciutils` is always pre-installed in RHCOS.

Do we want to complicate the script further for the maintainers by going the sysfs route?

-- Eli

tsorya: Hmm from my point it seems cleaner to say the truth and it will not require 
`[[ $bus_id != *":"*":"* ]] && full_id="0000:$bus_id" || full_id=$bus_id` as it will already normalized

eelgaev: Do we want to apply it for all mellanox devices and not only BF3 cards?

eelgaev:
Discovered a major problem, halting

tsorya:
closing, problem was in switchedev mode

PR #178, opened 2026-05-18
feat: add direct Redfish worker provisioning as alternative to BMO

eelgaev:
## Direct Redfish Worker Provisioning

Alternative to BMO/Ironic for provisioning physical servers as OCP workers. Uses Redfish VirtualMedia to mount the day2 ISO directly on the BMC and IPMI for reliable boot override.

### Why

BMO requires Ironic, Metal3, and a PXE/DHCP infrastructure. For environments with direct BMC access this is unnecessary overhead. Redfish VirtualMedia mounts the ISO directly — no network boot, no MAC address required.

### Flow

```
ISO download → VirtualMedia mount → IPMI boot CD → AI registration → bind/start → install → eject ISO → CSR approve → node Ready
```

### New env vars

| Variable | Description |
|----------|-------------|
| `WORKER_PROVISION_METHOD` | `bmo` (default) or `redfish` |
| `REDFISH_ISO_BASEURL` | HTTP base URL the BMC fetches the ISO from |
| `REDFISH_ISO_HOSTPATH` | Filesystem path on the HTTP server |
| `REDFISH_ISO_HOST` | SSH target for remote HTTP server (empty = local) |

Existing `WORKER_N_BMC_IP` is reused for IPMI. `WORKER_N_BOOT_MAC` not required.

### New files

- `scripts/redfish.sh` — Redfish API library (endpoint discovery, VirtualMedia, IPMI boot/power)
- `docs/user-guide/redfish-worker-provisioning.md` — setup guide (prerequisites, local vs remote HTTP, variable reference, troubleshooting)

### Also

- Makefile targets: `redfish-verify`, `redfish-power-status`

### Tested

End-to-end against Dell iDRAC R750 (Redfish 1.20.1) — server joined an SNO cluster as a worker node.

— Lyra

## Summary by CodeRabbit

* **New Features**
  * Optional direct Redfish worker provisioning alternative to BMO/Ironic, selectable via WORKER_PROVISION_METHOD.
  * New commands/Make targets to verify Redfish connectivity and check power status.
  * Support for serving day2 ISO over HTTP for Redfish virtual-media boot and automatic post-install labeling of DPU workers as worker-dpu.

* **Documentation**
  * Added a comprehensive Redfish Worker Provisioning user guide with setup and troubleshooting.


eelgaev:
Updated to integrate with Sharon's worker-dpu MachineConfigPool changes from #163.

Redfish provisioning bypasses BMO/MachineSet, so the `node-role.kubernetes.io/worker-dpu` label that the MachineSet normally assigns doesn't get set automatically. The latest commit adds a post-install labeling step so the worker-dpu MCP picks up Redfish-provisioned workers and applies all the DPU MachineConfigs (nmstate bridge, OVS mask, kernel args, etc.).

Also ensures the worker-dpu MCP exists for day-2 worker additions.

— Lyra 🤖

scripts/redfish.sh
```diff
@@ -0,0 +1,564 @@
...
+
+    local cmd="curl ${curl_flags} -X ${method} -u '${user}:${pass}' -H 'Content-Type: application/json'"
+
+    if [ -n "${body}" ]; then
+        cmd="${cmd} -d '${body}'"
+    fi
+
+    cmd="${cmd} '${url}'"
+    eval "${cmd}" 2>&1
+}
```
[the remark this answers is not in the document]
eelgaev: good catch bunny 🐰 — the eval was lazy and you're right it's a real risk.

Replaced the string-building + `eval` with a proper bash array (`CURL_FLAGS`) and direct `curl` invocation. No more credential leakage in `ps` output, no more injection risk from special chars in passwords.

— Lyra 🤖

scripts/worker.sh
```diff
@@ -338,6 +339,37 @@ _provision_workers_redfish() {
         # Don't fail — installation succeeded, node just needs time
     fi
 
+    # Step 9: Label worker nodes with worker-dpu role
```
tsorya, line 361: When you install worker-dpu you must install it with worker-dpu mcp pool and not with worker one, check how it should be done in aicli. Host while installing will get ignition from mc server and this ignition should be taken from worker-dpu pool. 
Before nodes joins the cluster you should pause worker mcp and unpause it only after host joined successfully

eelgaev: Good catch — fixed in e97c1a1.

Three changes:

1. **worker-dpu MCP ignition**: after `aicli bind`, we now call `aicli update host <id> -P mcp=worker-dpu` before starting installation. This sets `machine_config_pool_name` in the AI API so the host gets ignition rendered from the `worker-dpu` MachineConfigPool instead of the default `worker` pool. The DPU MachineConfigs (nmstate bridge, OVS mask, kernel args) are now baked into the initial ignition at install time.

2. **Pause worker MCP**: before nodes join the cluster, we `oc patch mcp worker --type merge -p '{"spec":{"paused":true}}'` to prevent MCO from rendering default worker configs on the new nodes before the `worker-dpu` label is applied.

3. **Unpause worker MCP**: after all nodes are labeled `worker-dpu`, we unpause the `worker` MCP. At that point the `worker-dpu` MCP selector matches the nodes so the default `worker` MCP won't claim them.

— Lyra 🤖

PR #183, opened 2026-05-26
No template management

omertuc:
Companion to https://github.com/rh-ecosystem-edge/dpf-hcp-provisioner-operator/pull/84

## Summary by CodeRabbit

* **New Features**
  * Added automatic generation of configuration overrides during post-install setup.
  * Added a new install command to create these overrides directly.

* **Bug Fixes**
  * Improved post-install manifest handling so generated settings are applied more consistently.
  * Updated service setup behavior for newer platform versions and feature combinations.


tsorya (approved):

josecastillolema:
I think this PR created a circular dependency that only manifests on a first-time/from-scratch install (upgrades are fine, since the template and DaemonSet already exist from a prior version):

1. Provisioner operator won't create DPUServiceTemplate/ovn until ovnkube-node-dpu-host DaemonSet rollout is complete.
2. That DaemonSet's pods can't schedule: they request the `openshift.io/bf3-p0-vfs-mgmt` SR-IOV VF resource, which is only advertised on a node once a DPU CR for it exists and reports `HostNetworkReady`.
3. DPU CRs are created by the DPUSet controller, which only runs after the parent DPUDeployment passes its dependency check,  which requires DPUServiceTemplate/ovn to already exist.

Net effect: no template → no DPU → no VF resource → DaemonSet stuck Pending forever → rollout never completes → template never created. DPUDeployment sits at PrerequisitesReady=False, worker nodes stay NotReady (missing CNI), indefinitely.

omertuc:
Possible, though "worked on my machine". Maybe something's changed or it's racey? Feel free to push any fixes, I only return on Thursday so won't be able to have a look until then

josecastillolema:
> Possible, though "worked on my machine". Maybe something's changed or it's racey? Feel free to push any fixes, I only return on Thursday so won't be able to have a look until then

Thanks @omertuc ! We'll work on some fixes in the meantime.
Just to confirm, when you sayed it worked on your machine, you mean a first time/from-scratch install?

omertuc:
> > Possible, though "worked on my machine". Maybe something's changed or it's racey? Feel free to push any fixes, I only return on Thursday so won't be able to have a look until then
> 
> Thanks @omertuc ! We'll work on some fixes in the meantime.
> Just to confirm, when you sayed it worked on your machine, you mean a first time/from-scratch install?

From scratch, yes

PR #188, opened 2026-06-08
[Automatic] Bump DPF Version to 26.4.0-e299bad8

eelgaev:
Use 26.4.0-e299bad8 by default

— Lyra 🤖


eelgaev:
Closing: upstream main already has DPF_VERSION=26.4.0-47c95221, which is newer than this PR's 26.4.0-e299bad8. This bump is no longer needed.

PR #207, opened 2026-06-24
Move ovnDaemonsetVersion to helm chart default with optional override + move to OCP 4.21 + remove feature gate

tsorya:
## Summary
- Removed hardcoded `ovnDaemonsetVersion` version-gating logic (1.1.0/1.2.0/1.3.0 based on OCP version) from `post-install.sh`
- Let the OVN helm chart provide its own default value for `ovnDaemonsetVersion`
- Added optional `OVN_DAEMONSET_VERSION` env var override: when set, it injects the value into the DPUServiceTemplate; when unset, the helm chart default applies
- Moved `ovnDaemonsetVersion` from `ovn-configuration.yaml` (DPUServiceConfiguration) to `ovn-template.yaml` (DPUServiceTemplate)
- Removed `99_feature-gate.yaml` (no longer needed for cluster installation)
- Updated default OpenShift version to 4.22.1

## Test plan
- [ ] Deploy with `OVN_DAEMONSET_VERSION` unset — verify helm chart default is used
- [ ] Deploy with `OVN_DAEMONSET_VERSION=1.3.0` — verify override is applied in generated manifest

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the OVN DaemonSet version via install and post-install workflows, including CI template substitution.
  * Updated CI defaults for OpenShift and OVN chart versions.
* **Bug Fixes**
  * Improved manifest generation so the OVN DaemonSet version is applied in the correct manifest stage.
  * Adjusted logging to reflect the updated version/MTU handling.
* **Breaking Changes**
  * Removed the OpenShift FeatureGate resource previously defined for the `cluster` name during cluster installation.


wabouhamad (approved):
/lgtm

PR #210, opened 2026-06-27
Add CronJob to complete stuck machine provisioning

josecastillolema:
When BMO workers use `customDeploy: install_coreos` with inspection disabled, BMHs have no `hardwareDetails` so the Machine controller can't match Nodes to Machines. Machines get stuck in Provisioned phase, firing `MachineWithNoRunningPhase` and `MachineWithoutValidNode` alerts.

This adds a CronJob that runs every 2 minutes on a control-plane node, resolves each worker's boot MAC via ARP (curl + ip neigh), and patches the node's `spec.providerID` to link it to its Machine. Gated behind `AUTO_COMPLETE_MACHINE_PROVISIONING=true`.

Tested on doca8: both worker machines transitioned from Provisioned to Running on the first CronJob run, alerts cleared, and subsequent runs were clean no-ops.

Assisted-by: Claude 🤖

## Summary by CodeRabbit

* **New Features**
  * Added optional automatic completion for worker machine provisioning.
  * Introduced a new scheduled cluster job to help finalize provisioning for stuck worker nodes.
  * Added new commands to deploy or remove the provisioning completer.

* **Chores**
  * Added a new environment setting to enable or disable automatic completion.
  * Updated configuration defaults and templates to include the new option.


josecastillolema:
@eelgaev @tsorya @SharonBX  perhaps there is a more elegant solution within the metal3 framework itself. Any reason why inspection is disabled for DPU workers? Does inspection not work with customDeploy, or does the DPU BMC doesn't support it?

josecastillolema:
Alredy fixed in https://github.com/rh-ecosystem-edge/openshift-dpf/pull/213

PR #211, opened 2026-06-29
fix: clear stale br-hbn bridge before recreating it in DPUFlavors

itailev:
The DPUFlavor rawConfigScript creates br-hbn with `add-br` but a stale br-hbn from a prior provisioning can retain leftover interfaces in OVS after the owning pod is deleted, causing connectivity issues on the DPU.

Tear down any existing br-hbn with `del-br --if-exists` immediately before recreating it, so each provisioning starts from a clean bridge. Applies to both the 1500 and 9000 MTU flavors.

Ref: NVBug 6345247

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network bridge setup by removing any existing `br-hbn` bridge before recreating it, helping avoid conflicts during post-installation configuration.
  * Made bridge reconfiguration more reliable across different DPU flavor setups.

### DIFF
--- a/manifests/post-installation/dpuflavor-1500.yaml
+++ b/manifests/post-installation/dpuflavor-1500.yaml
@@ -60,6 +60,7 @@ spec:
       _ovs-vsctl --may-exist add-br br-sfc
       _ovs-vsctl set bridge br-sfc datapath_type=netdev
       _ovs-vsctl set bridge br-sfc fail_mode=secure
+      _ovs-vsctl --if-exists del-br br-hbn
       _ovs-vsctl --may-exist add-br br-hbn
       _ovs-vsctl set bridge br-hbn datapath_type=netdev
       _ovs-vsctl set bridge br-hbn fail_mode=secure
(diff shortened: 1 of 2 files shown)


manifests/post-installation/dpuflavor-1500.yaml
```diff
@@ -60,6 +60,7 @@ spec:
       _ovs-vsctl --may-exist add-br br-sfc
       _ovs-vsctl set bridge br-sfc datapath_type=netdev
       _ovs-vsctl set bridge br-sfc fail_mode=secure
+      _ovs-vsctl --if-exists del-br br-hbn
       _ovs-vsctl --may-exist add-br br-hbn
```
souleb, line 64: so we do not expect it to exist here?

tsorya: It should exists only on redeploy

tsorya (approved):

manifests/post-installation/dpuflavor-1500.yaml
aserdean: it means we always delete and recreate the bridge during startup

PR #221, opened 2026-07-02
Upstream e2e testing integration

josecastillolema:
## Summary

Integrates upstream NVIDIA DPF e2e tests ([`NVIDIA/doca-platform/test/e2e`](https://github.com/NVIDIA/doca-platform/tree/public-main/test/e2e)) into this repo, allowing us to run the actual upstream Ginkgo test suite against a pre-existing DPF deployment without reimplementing any tests.

- Clones upstream repo at a pinned git ref, applies a small patch to skip `BeforeSuite` resource creation, and runs selected tests via Ginkgo label/focus filtering
- New wrapper script `scripts/e2e.sh` (setup/run/clean subcommands) and Makefile targets (`test-e2e`, `test-e2e-setup`, `test-e2e-clean`)
- Verified with the leader election failover test (`provisioning-controller`) — passed in ~36s against a live cluster

## Approaches considered

Three approaches were evaluated for integrating upstream e2e tests:

### Option A: Patch file (chosen)

Clone upstream at a pinned ref, apply a `git apply` patch to `e2e_suite_test.go` that adds a `DPF_SKIP_SETUP` env var guard to skip `BeforeSuite` resource creation, and run via `go test` with Ginkgo focus/label filtering.

| Pros | Cons |
|------|------|
| Runs the actual upstream test code — no reimplementation | Patch may need rebasing when upstream changes `e2e_suite_test.go` |
| Minimal diff (~30 lines): early return in `BeforeSuite` + nil guards for `cleanupTracker` | Tied to a specific upstream ref (though easy to bump) |
| Adding new tests = changing `E2E_FOCUS` / `E2E_LABEL_FILTER` env vars, no code changes | |
| Follows existing repo conventions (`scripts/`, `repos/`, Makefile targets) | |
| Upstream's `AfterSuite` skip flag (`-e2e.skip-cleanup.suite-after=true`) already exists — no patch needed there | |

> [!NOTE]
> The patch could be eliminated entirely if upstream added a native `--skip-setup` flag (analogous to the existing -`e2e.skip-cleanup.suite-after`), allowing BeforeSuite to be skipped without patching, **worth proposing upstream**.

### Option B: Fork/vendor the upstream test directory

Copy or git-subtree the entire `test/e2e/` directory into this repo, modify `BeforeSuite` directly, and maintain it as a vendored copy.

| Pros | Cons |
|------|------|
| Full control over test code | Must manually sync 40+ test files on every upstream update |
| No patch to maintain | Drift risk — hard to tell what's upstream vs. local |
| Can freely restructure tests | Large amount of Go code in a repo that is otherwise shell/YAML |

### Option C: Go module import (best approach, needs upstream changes)

Import the upstream test package as a Go module dependency (`go get github.com/nvidia/doca-platform/test/e2e`) and call exported test functions from a thin local test file.

| Pros | Cons |
|------|------|
| Clean separation — upstream is a dependency | Key types and helpers (`leaderElectionTarget`, `scaleControllerReplicas`, etc.) are ercase), blocking direct use |
| Automatic updates via `go get -u` | Adds a full Go module to a repo that is otherwise shell/YAML |
| No patch maintenance | |
| No drift risk — always consuming upstream code as-is | |

> [!NOTE]
> This option is currently blocked but could become the preferred approach if upstream exports the relevant types and helpers (e.g. `LeaderElectiplicas`). This would be a small non-breaking change: **worth proposing upstream**. If accepted, we could drop the patch entirely and consume the test library as a regular Go dependency.

## How it works

1. `make test-e2e-setup` shallow-clones `NVIDIA/doca-platform` at `DPF_UPSTREAM_TAG` (default: `public-main`) into `repos/doca-platform/` and applies `ci/patches/e2e-skip-setup.patch`
2. `make test-e2e` runs `go test ./test/e2e/` with:
   - `DPF_SKIP_SETUP=true` — triggers the patched early return in `BeforeSuite`
   - `DPF_E2E_COLLECT_RESOURCES=false` — disables resource collection on failure
   - `-e2e.skip-cleanup.suite-after=true` — prevents `AfterSuite` from deleting `DPFOperatorConfig`
   - `-ginkgo.focus` / `-ginkgo.label-filter` — selects which tests to run
   - Dummy values for `HELM_REGISTRY`, `TAG`, `NETUTILS_IMAGE`, `DOCKER_IO_REGISTRY` (required by `init()` but unused when setup is skipped)
3. `make test-e2e-clean` removes the cloned repo

### Configuration

All settings are overridable via environment variables:

| Variable | Default | Description |
|----------|---------|-------------|
| `DPF_UPSTREAM_TAG` | `public-main` | Upstream git ref to clone |
| `E2E_FOCUS` | `provisioning-controller leader pod is deleted` | Ginkgo focus regex |
| `E2E_LABEL_FILTER` | `DPFSystem` | Ginkgo label filter |
| `E2E_TIMEOUT` | `10m` | Test timeout |
| `E2E_KUBECONFIG` | `$KUBECONFIG` | Path to kubeconfig |

## Known limitations

- The `dpuservice-controller`, `kamaji-cluster-manager`, and `static-cluster-manager` leader election tests fail because these components are not exposed in `DPFOperatorConfig.ComponentConfigs()` for replica scaling — upstream gap, not a patch issue
- Currently pinned to `public-main` (HEAD) since the leader election test only exists there, not in release tags (e.g. `v26.4.0`). Once upstream cuts a release containing it, we should pin to a release tag for stability
- The patch needs to be verified/rebased when bumping `DPF_UPSTREAM_TAG`

## Test plan

- [x] `make test-e2e` runs the provisioning-controller leader election failover test and passes (~36s)
- [ ] Verify `make test-e2e-clean` removes the cloned repo
- [ ] Test with a different `E2E_FOCUS` to run additional upstream tests
- [ ] Pin to a release tag once upstream includes the leader election test

Assisted-by: Claude 🤖

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,8 @@ WORKER_SCRIPT := scripts/worker.sh
         delete-dpf-hcp-provisioner-operator \
         verify-deployment verify-workers verify-dpu-nodes verify-dpudeployment \
         run-traffic-flow-tests tft-setup tft-cleanup tft-show-config tft-results aicli-list \
-        validate-env-files generate-env deploy-observability
+        validate-env-files generate-env deploy-observability \
+        test-e2e test-e2e-setup test-e2e-clean
 
 all: 
 	@mkdir -p logs
@@ -227,6 +228,18 @@ run-dpf-sanity:
 	@chmod +x $(SANITY_CHECKS_SCRIPT)
 	@$(SANITY_CHECKS_SCRIPT)
 
+# E2E Tests (upstream NVIDIA DPF)
+E2E_SCRIPT := scripts/e2e.sh
+
+test-e2e-setup:
+	@$(E2E_SCRIPT) setup
+
+test-e2e: test-e2e-setup
+	@$(E2E_SCRIPT) run
+
+test-e2e-clean:
+	@$(E2E_SCRIPT) clean
+
 # Traffic Flow Tests
 run-traffic-flow-tests:
 	@echo "================================================================================"
@@ -367,6 +380,11 @@ help:
 	@echo "  verify-workers        - Wait for worker nodes to be Ready in host cluster"
 	@echo "  verify-dpu-nodes      - Wait for DPU nodes to be Ready in DPUCluster"
 	@echo "  verify-dpudeployment  - Wait for DPUDeployment to be Ready"
+	@echo ""
+	@echo "E2E Tests (upstream NVIDIA DPF):"
+	@echo "  test-e2e               - Setup and run upstream e2e tests (default: leader election)"
+	@echo "  test-e2e-setup         - Clone upstream repo and apply patch only"
+	@echo "  test-e2e-clean         - Remove cloned upstream repo"
 	@echo ""
 	@echo "Traffic Flow Tests:"
 	@echo "  run-traffic-flow-tests - Run kubernetes-traffic-flow-tests for network validation"
--- a/ci/patches/e2e-skip-setup.patch
+++ b/ci/patches/e2e-skip-setup.patch
@@ -1,13 +1,6 @@
 --- a/test/e2e/e2e_suite_test.go
 +++ b/test/e2e/e2e_suite_test.go
-@@ -1,5 +1,6 @@
- /*
- Copyright 2024 NVIDIA
-+Copyright 2026 Red Hat
-
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
-@@ -253,6 +254,12 @@ var _ = BeforeSuite(func() {
+@@ -253,6 +253,12 @@ var _ = BeforeSuite(func() {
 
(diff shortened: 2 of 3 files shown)


josecastillolema:
cc @linoyaslan @tsorya I personally like option C.

linoyaslan:
As I mentioned offline, I'm wondering how much of our test plan this approach can actually cover in practice.
Even where the upstream tests seem to overlap with our test plan, they're often not an exact match. That means that for a large portion of our test plan, we'll likely end up writing custom tests anyway, with the upstream code serving more as reference material than reusable infrastructure.

I wonder how you see us using this. if you have an example

tsorya:
I believe upstream test can be our framework

josecastillolema:
https://github.com/NVIDIA/doca-platform/pull/64 to follow approach C

PR #252, opened 2026-07-20
Enable secure boot for DPU deployment

shaharoni:
Added secure boot support for DPU deployment flow

## Summary by CodeRabbit

* **New Features**
  * Secure Boot is now enabled by default for DPU deployments.

### DIFF
--- a/manifests/post-installation/dpudeployment.yaml
+++ b/manifests/post-installation/dpudeployment.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: dpf-operator-system
 spec:
   dpus:
+    secureBoot: true
     nodeEffect:
       drain: true
     dpuSetStrategy:


manifests/post-installation/dpudeployment.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: dpf-operator-system
 spec:
   dpus:
+    secureBoot: true
```
tsorya, line 8: Will it require us to enable secure boot on all the dpus?
@eelgaev

eelgaev:
@shaharoni can you explain this change? On some clusters I can see secureboot in the status section of the dpu automatically.

manifests/post-installation/dpudeployment.yaml
eelgaev: I actually don't know what this spec setting does. 
Perhaps it affects their generated bf.cfg, which will have no effect here.

shaharoni:
this change is based on the assumption that we want spec.dpus.secureBoot: true in the DPUDeployment
if not, ignore it and I drop the change

eelgaev:
@shaharoni what happens if the DPU is not in secure boot mode?

shaharoni:
as I know, spec.dpus.secureBoot on DPUDeployment is the desired state
status.secureBoot.enabled on DPU / DPUDevice is the actual hardware state
when the spec field is set, DPF expects them to match

eelgaev:
I think we don't want to enforce secure boot state for DPUs at this point, some test clusters use signed drivers, some do not. 
I do believe we should document this in the installation doc, and it's a good idea to have it there optionally for security conscious users.

PR #263, opened 2026-07-31
Commented out the pings from hostnetwork pods

wabouhamad:
modified:   scripts/dpf-sanity-checks.sh

Commented out the ping tests from the workload pods using hostnetwork, since they were failing and not supported.


josecastillolema:
Humm you cannot test network PRs until we move the networking Prow test outside the bastion host. I'll start working on that.

PR #289, opened 2026-08-10
Discover TFT nodes from the cluster instead of BMH names

tsorya:
BMH/HBN hostnames are not Kubernetes node names, which caused TFT to fail when selecting server/client. Scan for Ready DPU workers and fall back to a master as client when only one DPU worker exists.

## Summary by CodeRabbit

- **New Features**
  - Traffic-flow tests now automatically discover Ready DPU worker nodes for server and client assignment.
  - Supports explicit server and client node overrides when needed.
  - Automatically selects a distinct Ready control-plane node as the client when only one DPU worker is available.
  - Resolves kubeconfig paths for more reliable test configuration.

- **Bug Fixes**
  - Prevents test execution when required nodes are missing or assigned identically.
  - Updated configuration guidance to reflect cluster-based node discovery.

### DIFF
--- a/ci/tft-config.yaml.template
+++ b/ci/tft-config.yaml.template
@@ -4,7 +4,7 @@
 # Repository: https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests
 #
 # This template is processed by scripts/traffic-flow-tests.sh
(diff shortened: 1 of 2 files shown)


wabouhamad (approved):
/lgtm

PR #294, opened 2026-08-12
E2E: enhance hosted cluster upgrade test with operational safeguards

tsorya:
## Summary

- Add pre-flight validation of the target release image via `oc adm release info` to catch typos/unreleased versions early
- Add `machineOSURL` handling: fail with guidance if set without `--upgrade-machine-os-url`, or patch/remove it before upgrade
- Verify ignition ConfigMap lifecycle during upgrade: deleted after upgrade starts, recreated with correct release image annotation (supports both old and new annotation keys)
- Add generation-aware DPU reprovisioning wait (`observedGeneration` gate to prevent false positives from stale status)
- Replace direct health check with retry-tolerant `waitForClusterHealth()` (15min tolerance for post-upgrade operator flapping)
- Remove hardcoded `Upgrading` phase assertion (compatible with [dpf-hcp-provisioner-operator#241](https://github.com/rh-ecosystem-edge/dpf-hcp-provisioner-operator/pull/241) phase renames)
- Add `Failed` phase detection during upgrade polling

Mirrors the operational safeguards from [#251](https://github.com/rh-ecosystem-edge/openshift-dpf/pull/251) (`upgrade-hosted.sh` / `upgrade-dpu.sh` scripts) into the e2e test.

## Test plan

- [ ] Run with `--upgrade-release-image` pointing to a valid z+1 release (e.g. `4.22.7-multi`)
- [ ] Verify test skips cleanly when `--upgrade-release-image` is not provided
- [ ] Verify `machineOSURL` step fails with guidance when set on cluster but flag not provided
- [ ] Verify ignition ConfigMap is validated during and after upgrade

## Summary by CodeRabbit

* **Tests**
  * Expanded hosted cluster upgrade coverage for machine OS URL updates and removals.
  * Added validation that upgrade configuration is properly deleted and recreated.
  * Improved checks for release image details, hardware reprovisioning, and overall cluster health.
  * Added support for configuring the machine OS upgrade endpoint in end-to-end test environments.
  * Strengthened upgrade verification across different release image annotation formats and provisioning states.

### DIFF
--- a/test/e2e/hosted_cluster_upgrade_test.go
+++ b/test/e2e/hosted_cluster_upgrade_test.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"fmt"
+	"os/exec"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -11,6 +12,8 @@ import (
 	provisioningv1 "github.com/nvidia/doca-platform/api/provisioning/v1alpha1"
 	dpfe2e "github.com/nvidia/doca-platform/test/e2e"
 
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -20,17 +23,14 @@ import (
 	"github.com/openshift-dpf/test/utils"
 )
 
-var dpfHCPProvisionerGVR = schema.GroupVersionResource{
-	Group:    "provisioning.dpu.hcp.io",
-	Version:  "v1alpha1",
-	Resource: "dpfhcpprovisioners",
-}
+const ignitionReleaseAnnotation = "dpfhcpprovisioner.dpu.hcp.io/bfcfg-template-ocp-release-image"
 
 var _ = Describe("Hosted Cluster Upgrade", Label("hosted-upgrade"), Ordered, func() {
 	var (
 		originalReleaseImage string
 		originalBFBName      string
 		newBFBName           string
+		dpuDeploymentGen     int64
 	)
 
 	BeforeAll(func() {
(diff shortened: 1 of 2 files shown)


test/e2e/hosted_cluster_upgrade_test.go
```diff
@@ -20,17 +23,17 @@ import (
...
 )
 
-var dpfHCPProvisionerGVR = schema.GroupVersionResource{
-	Group:    "provisioning.dpu.hcp.io",
-	Version:  "v1alpha1",
-	Resource: "dpfhcpprovisioners",
-}
+const (
+	ignitionReleaseAnnotationOld = "provisioning.dpu.nvidia.com/bfcfg-template-ocp-release-image"
+	ignitionReleaseAnnotationNew = "dpfhcpprovisioner.dpu.hcp.io/bfcfg-template-ocp-release-image"
```
linoyaslan: why we need both? not sure I get it

test/e2e/hosted_cluster_upgrade_test.go
```diff
@@ -65,18 +58,48 @@ var _ = Describe("Hosted Cluster Upgrade", Label("hosted-upgrade"), Ordered, fun
...
-			Kind:    "DPFHCPProvisioner",
-		})
+	It("should verify the target release image exists", func() {
+		By(fmt.Sprintf("Running oc adm release info for %s", cfg.UpgradeReleaseImage))
+		cmd := exec.CommandContext(ctx, "oc", "adm", "release", "info", cfg.UpgradeReleaseImage,
+			"-o", "jsonpath={.digest}")
+		output, err := cmd.CombinedOutput()
+		Expect(err).NotTo(HaveOccurred(),
+			"target release image %s not found or inaccessible: %s", cfg.UpgradeReleaseImage, string(output))
+		GinkgoWriter.Printf("Target image verified, digest: %s\n", string(output))
```
linoyaslan, line 65: do we really want to make the tc fail in that case? why not simply skip the tc? bc, it's not a real failure right? but just "can't test" - wdyt?

test/e2e/hosted_cluster_upgrade_test.go
tsorya: The operator changed this annotation key between versions — the old DOCA Platform annotation (`provisioning.dpu.nvidia.com/...`) was replaced by the dpf-hcp-provisioner annotation (`dpfhcpprovisioner.dpu.hcp.io/...`). We check both so the test works regardless of which operator version is deployed.

test/e2e/hosted_cluster_upgrade_test.go
tsorya: Good question! The intent here is to fail rather than skip — if someone explicitly provides an upgrade release image, pointing it at an invalid or inaccessible image is a configuration mistake that should be caught loudly. The skip path already exists in `BeforeAll` when the flag is not provided at all. So: empty image → skip (nothing to test), bad image → fail (misconfiguration).

PR #308, opened 2026-08-19
E2E: tc TC-MTU-003 change controlplaneMTU before deployment

linoyaslan:


test/e2e/mtu_controlplane_test.go
```diff
@@ -0,0 +1,257 @@
...
+		By(fmt.Sprintf("AfterAll: restoring controlPlaneMTU to %d", originalMTU))
+		configRestoreErr := patchControlPlaneMTU(originalMTU)
+		if configRestoreErr != nil {
+			GinkgoWriter.Printf("AfterAll: failed to restore controlPlaneMTU: %v\n", configRestoreErr)
+		}
+
+		// The operator reads controlPlaneMTU while reconciling the
+		// DPUDeployment. Recreate it after restoring the config so the
+		// original MTU is propagated to the bridges as well.
+		restoreDPUDeploymentFromBackup(dpuDeploymentBackup)
```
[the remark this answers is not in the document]
linoyaslan: @josecastillolema / @tsorya I intentionally removed it because I think it's too much to wait twice - once when the test runs, and then again after restoring the original state. WDYT?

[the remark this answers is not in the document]
josecastillolema: Did you remove the validation before starting the test of after the test?
I think we should keep both ...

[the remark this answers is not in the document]
linoyaslan: At first, I added `waitForClusterHealthAfterDPUReprovisioning` to the `AfterAll` but then I removed it because it seemed unnecessary to me. The test itself has already passed at that point, and the `AfterAll` is only restoring the initial configuration, so why do we need to wait for the cluster to become healthy again? If something goes wrong during the restore, or the health check reaches a timeout for example, why should that cause the test itself to fail? From my point of view, the actual test passed, so reporting it as failed because of the restore step could be a bit misleading.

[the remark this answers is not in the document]
josecastillolema: But we are chaining tests right from a CI perspective right?
After e2e we may have scale, network or others.
The assumption for e2e is that cluster must be healthy before and after the test so we can safely chain tests.

test/e2e/mtu_controlplane_test.go
```diff
@@ -0,0 +1,299 @@
...
+		return fmt.Errorf("restoring DPFOperatorConfig controlPlaneMTU: %w", err)
+	}
+	return nil
+}
+
+// toggleMTU returns a different valid MTU value than current: 1500 if
+// current is 9000, otherwise 9000. Mirrors the two NODES_MTU values
+// supported by this repo's deployment scripts (scripts/env.sh validate_mtu).
+func toggleMTU(current int) int {
+	if current == 9000 {
```
tsorya: How it should work, if current control plane mtu is 9000 how we can change it to 1500? how it will work

test/e2e/helpers.go
```diff
@@ -110,3 +126,96 @@ func discoverWorkloadPods(ctx context.Context, c client.Client, namespace string
 
 	return result, nil
 }
+
+// findOVNNodePodOnNode returns a ready OVN node pod scheduled on nodeName.
+func findOVNNodePodOnNode(nodeName string) (*corev1.Pod, error) {
```
tsorya:
```suggestion
func findOVNNodePodOnNode(nodeName string) (*corev1.Pod, error) {

func findOVNKNodePodOnNode(nodeName string) (*corev1.Pod, error) {

```

test/e2e/helpers.go
```diff
@@ -110,3 +126,96 @@ func discoverWorkloadPods(ctx context.Context, c client.Client, namespace string
...
+	podList := &corev1.PodList{}
+	if err := hostedClient.List(ctx, podList,
+		client.InNamespace(cfg.DPFNamespace),
+		client.MatchingLabels{ovnNodeComponentLabel: ovnNodeComponentLabelValue},
+	); err != nil {
+		return nil, fmt.Errorf("listing OVN node pods in %s: %w", cfg.DPFNamespace, err)
+	}
+	for i := range podList.Items {
+		if podList.Items[i].Spec.NodeName == nodeName &&
+			len(ovnKPodReadinessProblems(&podList.Items[i])) == 0 {
```
tsorya: hmm , can you elaborate what we are doing here?

test/e2e/helpers.go
```diff
@@ -110,3 +126,96 @@ func discoverWorkloadPods(ctx context.Context, c client.Client, namespace string
...
+		}
+	}
+	return nil, nil
+}
+
+// getInterfaceMTU executes `ip link show <iface>` in a pod container and
+// parses the reported MTU.
+func getInterfaceMTU(ctx context.Context, restCfg *rest.Config, cs *kubernetes.Clientset, namespace, podName, containerName, iface string) (int, error) {
+	result, err := utils.ExecInPod(ctx, restCfg, cs, namespace, podName, containerName, []string{
+		"ip", "link", "show", iface,
```
tsorya: ip -j link show iface | jq '.[0].mtu'

or get json and parse it in go

test/e2e/helpers.go
```diff
@@ -110,3 +126,96 @@ func discoverWorkloadPods(ctx context.Context, c client.Client, namespace string
...
+		g.Expect(apierrors.IsNotFound(err)).To(BeTrue(),
+			"DPUDeployment should be fully deleted, got: %v", err)
+	}).WithTimeout(10 * time.Minute).WithPolling(5 * time.Second).Should(Succeed())
+
+	GinkgoWriter.Printf("DPUDeployment %s deleted successfully\n", cfg.DPUDeploymentName)
+}
+
+// recreateDPUDeploymentFromBackup creates a new DPUDeployment using the
+// ObjectMeta and Spec captured from a previous deployment.
+func recreateDPUDeploymentFromBackup(backup *dpuservicev1.DPUDeployment) {
```
tsorya: WHy not just create dpu deployment? I mean why we need a dedicated function for creating a dpu deployment from the object?

test/e2e/mtu_controlplane_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// defaultControlPlaneMTU mirrors the CRD's +kubebuilder:default for
+	// Networking.ControlPlaneMTU, used when the field is unset.
+	defaultControlPlaneMTU = 1500
+)
+
+// The host-side OOB bridge is named br-dpu by the DPF deployment manifests,
+// while x86/OVN host configurations commonly expose the same network as
+// br-ex. Probe both names and use the one carrying the expected MTU.
+var hostOOBBridgeNames = []string{"br-dpu", "br-ex"}
```
tsorya: we moved to "br-ex"

test/e2e/mtu_controlplane_test.go
```diff
@@ -0,0 +1,299 @@
...
+		// original MTU is propagated to the bridges as well.
+		restoreDPUDeploymentFromBackup(dpuDeploymentBackup)
+		// Do not run the full DPU/OVN health gate here. Subsequent tests perform their
+		// normal reprovisioning preflight before making further changes.
+		Expect(configRestoreErr).NotTo(HaveOccurred(), "AfterAll: restoring controlPlaneMTU")
+	})
+
+	// ── Pre-conditions ──────────────────────────────────────────────────────
+
+	It("pre-condition: should have a DPUDeployment in Ready state", func() {
```
tsorya: should not it be part of before all?

test/e2e/mtu_controlplane_test.go
linoyaslan: I know that's why I added br-ex bc before that it was br-dpu. unless you want me to remove br-dpu entirely?

tsorya: I believe there is no point in br-dpu anymore

tsorya (approved):

test/e2e/helpers.go
linoyaslan: I don't understand the suggestion

test/e2e/helpers.go
linoyaslan: looks for a ready OVN node pod in the configured DPF namespace that is scheduled on nodeName.
It:
  1. Lists pods in DPFNamespace with the OVN node component label.
  2. Checks each pod's Spec.NodeName.
  3. Uses ovnKPodReadinessProblems to decide whether that pod is ready.
  4. Returns the first pod that matches both checks.

test/e2e/helpers.go
linoyaslan: will fix on 2nd PR as this one already merged

test/e2e/helpers.go
linoyaslan: bc we want to restore the cluster to its original state. In my mind, if you run a test on your cluster, you'd expect the cluster to return to its original state once the test is complete. or do you see it differently?

test/e2e/mtu_controlplane_test.go
linoyaslan: ok. let me fix it in 2nd PR as this one already merged

test/e2e/mtu_controlplane_test.go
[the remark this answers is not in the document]
linoyaslan: ok sure np. let me fix it on a 2nd PR as this one already merged

test/e2e/mtu_controlplane_test.go
linoyaslan: you are right, it actually already part of skipIfClusterNotReadyForDPUReprovisioning
let me remove this It in 2nd PR as this one already merged

test/e2e/mtu_controlplane_test.go
linoyaslan: I don't understand the q

PR #313, opened 2026-08-19
Skip alt-arch InfraEnv creation for non-multi-arch clusters

tsorya:
When the cluster is not multi-arch (OPENSHIFT_VERSION lacks the -multi suffix), creating an InfraEnv for the alternative architecture fails with a 400 Bad Request from the Assisted Installer because the CPU architecture doesn't match the cluster.

Guard the alt-arch InfraEnv creation behind is_multi_arch_version() so single-arch clusters can still use the day2 flow for same-arch workers.

### DIFF
--- a/scripts/cluster.sh
+++ b/scripts/cluster.sh
@@ -669,6 +669,14 @@ function deploy_odf() {
 # ISO management functions
 # -----------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


eelgaev (approved):
/lgtm

PR #328, opened 2026-08-24
fix: use dpuNodeSelector, drop redundant BFB fileName

itailev:
- DPUDeployment dpuSets.nodeSelector is deprecated and removed in DPF v26.7.0; dpuNodeSelector is an equivalent rename. 
- BFB spec.fileName is optional and DPF derives a unique <namespace>-<name>.bfb when unset.

Both align with upstream DPF conventions and the DPF-on-OpenShift user manual. No behavioral change.


itailev:
we can merge this only to main, no urgency in merging into 4.22

tsorya (approved):

PR #331, opened 2026-08-25
Sanity: replace awk parsing with jsonpath

josecastillolema:
Replaced fragile awk column parsing with jsonpath for extracting hosted cluster kubeconfig, and added validation to catch empty results early.

Latest failure on the daily run: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-rh-ecosystem-edge-openshift-dpf-main-daily-network-doca4/2092047121019572224
```
 Output of oc get nodes --kubeconfig='/root/XXXXX/ci/openshift-dpf-2026-08-25_00-51-12/openshift-dpf/doca-kubeconfig-2026-08-24_22-59-40':
error: Missing or incomplete configuration info.  Please point to an existing, complete config file:
```

### DIFF
--- a/scripts/dpf-sanity-checks.sh
+++ b/scripts/dpf-sanity-checks.sh
@@ -50,21 +50,19 @@ echo -e "- SANITY_TESTS_PING_HBN_TO_HBN_PODS: '${SANITY_TESTS_PING_HBN_TO_HBN_PO
 mgmt_kubecfg="${KUBECONFIG}"
 echo -e "\n- mgmt_kubecfg: '${mgmt_kubecfg}'"
 
-# Get the hosted cluster kubeconfig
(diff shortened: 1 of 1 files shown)


schmaustech:
I would be interested in understanding why awk is fragile - is that based on the premise the output is going to radically change?   Either method works but this feels like all we did was swap out a filament light bulb with and led to get merge count.

josecastillolema:
@schmaustech correct me if I am wrong but the jsonpath approach is generally better, it queries the API object directly, so you get exactly the field without relying on column positions or text formatting.
I have a strong suspect that the previous approach made this [test](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-rh-ecosystem-edge-openshift-dpf-main-daily-network-doca4/2092047121019572224) fail and this new approach won't.
I do not understand the comment about the merge count.

PR #332, opened 2026-08-25
New test 2nd interface two rp

SharonBX:


scripts/traffic-flow-tests.sh
```diff
@@ -42,6 +42,23 @@ TFT_TEST_CASES="${TFT_TEST_CASES:-1-25}"
 TFT_DURATION="${TFT_DURATION:-10}"
```
josecastillolema, line 42: Dont you need to add new test cases?

SharonBX: add it and gate it only if we have dpu-host

josecastillolema: humm tft is a project specific to DPU enabled clusters, I dont think we need that gate and we can have it enabled by default, wdyt?

SharonBX: you are right but soon cluster will be also mixed with regular workers (no dpu)

josecastillolema: So this gate is generic or specific to this use case? If it is specific to this use case I suggest we drop it. When in the future we have scenarios with mixed workers and workers-dpu let's just filet for workers-dpu nodes for all the test cases.

SharonBX: its totally generic, this TC will run when he find dpu worker in cluster, why remove this flexibility ?

josecastillolema: Can you point me to this gate?
Its not about removing flexibility, its about making sure we dont duplicate logic in different places. We already have a logic for choosing the appropiate workers, we dont need additionals gates IMO

SharonBX: got you ... 
https://github.com/rh-ecosystem-edge/openshift-dpf/pull/332/changes/f345521b0f8a2f7666aff479bc80aee90ba1965a#diff-b977e47f6fca127097a0ca20196d3e1100109921d415ea71e1ca773b1d166e1aR378

PR #340, opened 2026-08-26
fix: correct OVN pod network MTU for jumbo frames

itailev:
Found  as part of the GA User Manual review (and alignment against this automation repo).

The existing branches use inconsistent encapsulation overheads: the 1500-byte profile maps to 1400, while the 9000-byte profile maps to 8940.

Use OVN-Kubernetes' 100-byte Geneve overhead for both supported MTU profiles. This preserves 1400 for the default and produces 8900 for jumbo frames.


itailev:
@tsorya FYI

tsorya (approved):

PR #347, opened 2026-08-27
Install OVN resource injector as a Helm release

tsorya:
## Summary
- Replace `helm template` + `oc apply` with `helm upgrade --install` so the OVN resource injector is a real Helm release (`ovn-kubernetes` in `openshift-ovn-kubernetes`).
- Chart URL, version, resource name, and webhook ports stay env-driven; other OVN chart manifests remain disabled.
- Lets later injector chart upgrades take over the same objects instead of leaving them unmanaged.

## Test plan
- [ ] `make enable-ovn-injector` on a cluster that does not yet have the injector: Helm release is created, webhook deployment rolls out, MutatingWebhookConfiguration and `dpf-ovn-kubernetes` NAD exist.
- [ ] Re-run `make enable-ovn-injector`: `helm upgrade --install` succeeds (idempotent).
- [ ] On a cluster that still has injector objects from the old `oc apply` path: confirm whether Helm can adopt them, or if a one-time delete/recreate is needed (this change does not pass `--take-ownership`).

### DIFF
--- a/scripts/enable-ovn-injector.sh
+++ b/scripts/enable-ovn-injector.sh
@@ -23,18 +23,11 @@ INJECTOR_WEBHOOK_PORT=19443
 INJECTOR_HEALTH_PROBE_PORT=18081
 INJECTOR_METRICS_PORT=29091
 
-log [INFO] "Enabling OVN resource injector..."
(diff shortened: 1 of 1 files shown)


wabouhamad (approved):
/lgtm

PR #371, opened 2026-09-02
Pr2 tft secondary interface

SharonBX:


manifests/post-installation/nodesriovdevicepluginconfig.yaml
```diff
@@ -19,6 +19,11 @@ spec:
         - pfIndex: 0
           start: 2
           end: <NUM_VFS_END>
+    - name: <SRIOV_DP_SECONDARY_POOL_NAME>
+      type: vf
+      options:
+        isRdma: true
+      ranges:
         - pfIndex: 1
           start: 0
```
tsorya, line 23: Lets not use here all the pool for secondary network, just take 5 vf for it

PR #372, opened 2026-09-03
Gate DPUServiceTemplate override generation behind a flag

tsorya:
## Summary
- Add `GENERATE_DPUSERVICETEMPLATE_OVERRIDES` (default `false`) so `prepare-dpu-files` / `deploy-dpu-services` / `make all` no longer always write the `dpuservicetemplate-overrides` ConfigMap.
- Set the flag to `true` to opt in during prepare, when custom OVN/HBN/DTS chart or image values should be applied.
- `make generate-overrides` still writes the ConfigMap explicitly, independent of the flag.

## Test plan
- [ ] `make validate-env-files` succeeds (new var is present in both `ci/env.defaults` and `ci/env.template`)
- [ ] With default `.env`, `make prepare-dpu-files` logs that DPUServiceTemplate overrides were skipped and does not create/update `dpuservicetemplate-overrides`
- [ ] With `GENERATE_DPUSERVICETEMPLATE_OVERRIDES=true`, `make prepare-dpu-files` creates the ConfigMap in `DPF_HCP_PROVISIONER_OPERATOR_NAMESPACE`
- [ ] `make generate-overrides` still creates the ConfigMap when the flag is unset/false


eelgaev (approved):
/lgtm

PR #374, opened 2026-09-03
[release-4.22] Gate DPUServiceTemplate override generation behind a flag

tsorya:
## Summary
Cherry-pick of #372 onto `release-4.22`.

- Add `GENERATE_DPUSERVICETEMPLATE_OVERRIDES` (default `false`) so `prepare-dpu-files` / `deploy-dpu-services` / `make all` no longer always write the `dpuservicetemplate-overrides` ConfigMap.
- Set the flag to `true` to opt in during prepare, when custom OVN/HBN/DTS chart or image values should be applied.
- `make generate-overrides` still writes the ConfigMap explicitly, independent of the flag.

## Test plan
- [ ] `make validate-env-files` succeeds
- [ ] With default `.env`, `make prepare-dpu-files` skips DPUServiceTemplate overrides
- [ ] With `GENERATE_DPUSERVICETEMPLATE_OVERRIDES=true`, `make prepare-dpu-files` creates the ConfigMap
- [ ] `make generate-overrides` still creates the ConfigMap when the flag is unset/false


eelgaev (approved):
/lgtm

PR #377, opened 2026-09-04
Add Kata DPU cold-plug support with scalable test Deployment

tsorya:
## Summary
- Add optional Kata DPU cold-plug (not part of `make all`): OSC, worker-dpu MachineConfigs (RHCOS layer, IOMMU, CRI-O kata-coldplug handler), RuntimeClass `kata-coldplug`, shared VF pool, and OVN injector kata NAD mapping when `KATA_ENABLED=true`.
- OSC is installed because the Kata team required it for this path. DPU hosts already live in MCP `worker-dpu`, and MCO allows only one custom pool per node, so OSC must not add `node-role.kubernetes.io/kata-oc` on those hosts. KataConfig uses a selector that matches no nodes (`openshift-dpf.io/kata-oc-do-not-select`) so OSC stays installed without moving DPU workers.
- Cold-plug MachineConfigs and RuntimeClass `kata-coldplug` target `worker-dpu`. Test workload is `kata-dpu-test` (netshoot); scale with `KATA_TEST_REPLICAS` via `make deploy-kata-test`.

## Test plan
- [ ] `KATA_ENABLED=true make enable-ovn-injector` then `make enable-kata` installs OSC, inert KataConfig, worker-dpu MCs, and RuntimeClass without labeling DPU nodes `kata-oc`
- [ ] `make deploy-kata-test` (default 1 replica) reaches Available; `oc exec deploy/kata-dpu-test -- ping -c 3 8.8.8.8` works
- [ ] `KATA_TEST_REPLICAS=4 make deploy-kata-test` scales without oversubscribing VFs (`maxSurge: 0`)
- [ ] Replicas do not exit after 5 minutes (sleep infinity)
- [ ] `make validate-env-files` succeeds (`KATA_ENABLED` / `KATA_TEST_REPLICAS` in defaults and template)


ci/env.defaults
```diff
@@ -23,9 +23,9 @@ SSH_KEY=${SSH_KEY:-~/.ssh/id_ed25519.pub}
 # DPF Configuration
 DPF_VERSION=${DPF_VERSION:-v26.4.1}
 DPF_HELM_REPO_URL=${DPF_HELM_REPO_URL:-https://helm.ngc.nvidia.com/nvidia/doca}
-OVN_CHART_URL=${OVN_CHART_URL:-oci://ghcr.io/mellanox/charts}
-OVN_TEMPLATE_CHART_URL=${OVN_TEMPLATE_CHART_URL:-oci://ghcr.io/mellanox/charts}
-OVN_CHART_VERSION=${OVN_CHART_VERSION:-v26.4.1-ocp-release-v4.22}
+OVN_CHART_URL=${OVN_CHART_URL:-oci://quay.io/itsoiref/charts}
```
souleb: why do we need your fork? Everything should be in the mellanox fork already

docs/kata-dpu-coldplug.md
```diff
@@ -0,0 +1,391 @@
...
+KATA_SRIOV_DP_CONFIG_NAME=bf3-p1-vfs-kata
+KATA_SRIOV_PF_INDEX=1
+KATA_NUM_VFS=24            # kata pods on PF1; regular gets NUM_VFS - KATA_NUM_VFS
+KATA_RUNTIME_CLASS=kata-coldplug
+KATA_INJECTOR_RESOURCE_NAME=openshift.io/bf3-p1-vfs-kata
+# Git default: dpf-ovn-kubernetes-${KATA_RUNTIME_CLASS}
+# Override example seen on nvd-srv-45:
+KATA_NAD_NAME=dpf-ovn-kubernetes-bf3-p1-vfs-kata
+KATA_RHCOS_LAYER_IMAGE=quay.io/jensfr/rhcos-kata-dpu@sha256:ce05dea3e0214c7bf7864cef1110e414a6419430fe2f55b5e9c848b71b799a8f
+KATA_SKIP_RHCOS_LAYER=false
```
souleb, line 42: so we don't have the zstream testable at the moment?

manifests/kata/06-test-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+        labelSelector:
+          matchLabels:
+            app: kata-dpu-test
+      containers:
+      - name: test
+        image: quay.io/wabouham/ecosys-nvidia/nicolaka-netshoot:latest
+        command: ["sleep", "infinity"]
+        resources:
+          limits:
+            "<KATA_INJECTOR_RESOURCE_NAME>": "1"
```
souleb: no needed, if the runtime class is matched, the injector will add 1 resource

ci/env.defaults
tsorya: Yep, mistake, will change it.

docs/kata-dpu-coldplug.md
tsorya: Nope we need to apply a layer with a new rpm

scripts/post-install.sh
```diff
@@ -211,14 +211,31 @@ function prepare_post_installation() {
...
         return 1
     fi
+    if ! [[ "${NUM_VFS}" =~ ^[1-9][0-9]*$ && "${KATA_NUM_VFS}" =~ ^[1-9][0-9]*$ ]]; then
+        log [ERROR] "NUM_VFS and KATA_NUM_VFS must be positive integers"
+        return 1
+    fi
+    if [ "${KATA_NUM_VFS}" -ge "${NUM_VFS}" ]; then
+        log [ERROR] "KATA_NUM_VFS (${KATA_NUM_VFS}) must be less than NUM_VFS (${NUM_VFS})"
+        return 1
+    fi
```
itailev: `^[1-9][0-9]*$` rejects empty, and this runs unconditionally in a function reached from
`_all` via `prepare-dpu-files`. Existing clusters have no `KATA_NUM_VFS` in `.env` —
`make generate-env` refuses to overwrite without `FORCE=true`, and the scripts use
`set -e` but not `set -u`, so it expands empty. `make all` then aborts with "NUM_VFS and
KATA_NUM_VFS must be positive integers" on every pre-existing deployment. Defaulting
`KATA_NUM_VFS` in `env.sh` as well would fix it.

It also rejects `0`, so no value disables Kata. Accepting `0` and skipping the pool
render would — same as CodeRabbit's open comment on `ci/env.defaults:62`.

ci/env.defaults
```diff
@@ -53,8 +53,28 @@ SRIOV_DP_MGMT_POOL_NAME=${SRIOV_DP_MGMT_POOL_NAME:-bf3-p0-vfs-mgmt}
 INJECTOR_RESOURCE_NAME=${INJECTOR_RESOURCE_NAME:-${SRIOV_DP_RESOURCE_PREFIX}/${SRIOV_DP_CONFIG_NAME}}
 INJECTOR_CHART_VERSION=${INJECTOR_CHART_VERSION:-${OVN_CHART_VERSION}}
 OVNK_NAMESPACE=${OVNK_NAMESPACE:-openshift-ovn-kubernetes}
+# Total VFs per PF. PF1 regular pool size is NUM_VFS - KATA_NUM_VFS (- other pools later).
 NUM_VFS=${NUM_VFS:-46}
 
+# Kata Containers Configuration
+KATA_SRIOV_DP_CONFIG_NAME=${KATA_SRIOV_DP_CONFIG_NAME:-bf3-p1-vfs-kata}
+KATA_SRIOV_PF_INDEX=${KATA_SRIOV_PF_INDEX:-1}
+KATA_NUM_VFS=${KATA_NUM_VFS:-24}  # PF1 kata VFs; regular PF1 pool gets NUM_VFS - KATA_NUM_VFS
```
itailev: Suggest `KATA_NUM_VFS=${KATA_NUM_VFS:-0}`. At 24, `prepare-dpu-files` (in `_all`) halves
the PF1 regular pool on every deployment, including ones that never run
`make enable-kata`. Needs the validation on `scripts/post-install.sh` to accept 0.

scripts/enable-ovn-injector.sh
```diff
@@ -70,4 +73,12 @@ else
...
 
-log [INFO] "OVN resource injector enabled successfully"
+# Verify Kata NAD creation
+if oc get net-attach-def -n "${OVNK_NAMESPACE}" "${KATA_NAD_NAME}" &>/dev/null; then
+    log [INFO] "NetworkAttachmentDefinition '${KATA_NAD_NAME}' created successfully"
+else
+    log [ERROR] "NetworkAttachmentDefinition '${KATA_NAD_NAME}' was not created"
+    exit 1
+fi
+
```
itailev: This `exit 1` runs on a target in `_all`, so a deployment that never enables Kata fails
on a missing NAD it does not need. Suggest gating on an explicit `KATA_ENABLED` and
warning instead of exiting.

manifests/post-installation/nodesriovdevicepluginconfig.yaml
```diff
@@ -21,4 +21,12 @@ spec:
...
-          end: <NUM_VFS_END>
+          end: <PF1_REGULAR_VF_END>
+    - name: <KATA_SRIOV_DP_CONFIG_NAME>
+      type: vf
+      options:
+        isRdma: true
+      ranges:
+        - pfIndex: <KATA_SRIOV_PF_INDEX>
+          start: <KATA_VF_START>
+          end: <KATA_VF_END>
```
itailev: **Drop this pool.** `options.isRdma` is the only per-pool knob that can differ between
container and Kata VFs, and it is a no-op for Kata — the guest binds its own `mlx5_ib` to
the passed-through function. One pool with `isRdma: true` serves both; `runtimeClassName`
still selects the runtime and the injector still supplies the per-runtime NAD. That is a
revert of line 24 to `<NUM_VFS_END>` plus deleting lines 25-32:

```yaml
    - name: <SRIOV_DP_CONFIG_NAME>
      type: vf
      options:
        isRdma: true
      ranges:
        - pfIndex: 0
          start: 2
          end: <NUM_VFS_END>
        - pfIndex: 1
          start: 0
          end: <NUM_VFS_END>
```

It also removes the PF1-only exposure: as written every Kata VM rides a single uplink, so
losing that port drops all Kata workloads while PF0 pods keep running.

`KATA_SRIOV_DP_CONFIG_NAME`, `KATA_SRIOV_PF_INDEX`, `KATA_VF_START`, `KATA_VF_END` and
`PF1_REGULAR_VF_END` become dead substitutions in `scripts/post-install.sh`.

Verified that the injector permits a shared pool. `RuntimeClassNADMappings` is a
`map[string]string` of runtimeClass to NAD name only — the chart passes just
`--runtime-class-nad-mapping=<class>=<nadName>` — and `getVFResourceName` then reads the
resource from that NAD's `k8s.v1.cni.cncf.io/resourceName` annotation at admission
(`dpf-utils/internal/ovnkubernetesresourceinjector/webhooks/webhook.go` in
`Mellanox/ovn-kubernetes-dpf`). Confirmed on an OCP 4.22 DPF cluster: the deployed
injector carries no `--resource-name` arg, and a server-side dry-run pod came back with
`openshift.io/bf3_vfs: "1"` injected straight from the NAD annotation.

So `resourceName` can be shared across mappings — only `nadName` must stay unique, since
`nad.yaml` renders one NAD per mapping.

manifests/kata/02-kata-machineconfig.yaml
```diff
@@ -0,0 +1,38 @@
...
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: <WORKER_ROLE>
+  name: 99-kata-dpu
+spec:
+  <KATA_OS_IMAGE_URL>
+  kernelArguments:
+    - intel_iommu=on
+    - iommu=pt
```
itailev: 1. `intel_iommu=on` only — add `amd_iommu=on` so this is not a silent no-op on AMD DPU
   hosts.
2. This MachineConfig carries both `osImageURL` and `kernelArguments`. The jensfr README
   warns that on image-layered RHCOS `kernelArguments` may not apply, and splits them
   into `03-rhcos-layer.yaml` plus a separate `99-iommu-enable`. Has the karg been
   confirmed to land with `KATA_SKIP_RHCOS_LAYER=false`? `cat /proc/cmdline | grep iommu`
   on a rolled-out node settles it.

scripts/enable-kata.sh
```diff
@@ -0,0 +1,272 @@
...
+    fi
+
+    local timeout=$((KATA_TEST_REPLICAS * 180))
+    log [INFO] "Waiting for deployment/kata-dpu-test (${KATA_TEST_REPLICAS} replicas, timeout ${timeout}s)..."
+    oc wait --for=condition=Available deployment/kata-dpu-test --timeout="${timeout}s"
+    oc get pods -l app=kata-dpu-test -o wide
+    log [INFO] "Verify one replica: oc exec deploy/kata-dpu-test -- ping -c 3 8.8.8.8"
+}
+
+function enable_kata() {
```
itailev, line 360: `wait_for_mcp` returns as soon as the MCP reports updated, but the runbook documents that
a host reboot leaves the DPU without an IPv4 on `br-dpu` and breaks all pod networking.
Since this path always reboots, it reports success while networking is still down.

Also cheap: an `ls /dev/kvm` check before installing OSC and rebooting would turn the
`VMX disabled by BIOS` case from the runbook into an immediate error rather than a wasted
reboot cycle.

itailev (changes requested):
Requesting changes. Three changes take effect on the default `make all` path even when
Kata is never enabled:

1. The new `KATA_NUM_VFS` validation rejects empty and runs unconditionally in
   `prepare_post_installation`, which is in `_all` — so `make all` now aborts on any
   cluster whose `.env` predates this PR.
2. The PF1 repartition is unconditional: `bf3-p0-vfs` drops from 90 VFs to 66 on
   clusters that never enable Kata.
3. The NAD check hard-exits on a target that is also in `_all`.

Separately, the dedicated Kata VF pool may not be needed at all — `options.isRdma` is the
only per-pool knob that can differ between container and Kata VFs, and it is a no-op for
Kata. Noted inline.

One question: with no KataConfig created, what is the OSC subscription for? `ensure_osc`
blocks up to 10 minutes on CSV Succeeded, but the Kata bits come from the RHCOS layer.

scripts/enable-kata.sh
tsorya: > wait_for_mcp returns as soon as the MCP reports updated, but the runbook documents that
> a host reboot leaves the DPU without an IPv4 on br-dpu and breaks all pod networking.
> Since this path always reboots, it reports success while networking is still down.
No need to add complexity for it

manifests/post-installation/nodesriovdevicepluginconfig.yaml
tsorya: No, we can't.
By default we are setting isRdma: true
kata doesn't support it and will remove the setting from VF and nothing will add it back after it will free it

itailev:
@tsorya 2 more minor issues:
1. `scripts/env.sh:141` skips quote-stripping under `make`. I used `KATA_ENABLED="true"`  by mistake and it silently disabled the whole kata path.
2. `ci/env.defaults:73-74` says MCO "will fail / drift" on RHCOS-layer mismatch, but on my setup it just downgraded both workers. setup was up (no MCO failure)

PR #387, opened 2026-09-07
Backport maintenance operator config deploy 4.22

itailev:


itailev:
targeted the wrong base,

PR #398, opened 2026-09-16
NVIDIA-1335: Run custom post-install script at end of make all

eelgaev:
## What

Adds `CUSTOM_POSTINSTALL_SCRIPT`: a single, opt-in parameter that runs a user-supplied script as the **last** step of `make all`.

```
# Local file, with arguments
CUSTOM_POSTINSTALL_SCRIPT="/home/user/bla/thescript.sh arg1 arg2"

# http(s) URL, with arguments
CUSTOM_POSTINSTALL_SCRIPT="https://example.com/thescript.sh arg1 arg2"
```

## How

- New `scripts/custom-postinstall-script.sh` (`run` subcommand) parses the value as `<script> [args...]`: the first token is the script, the rest are passed to it as arguments.
  - A local file path is run directly (errors if missing).
  - An `http(s)` URL is downloaded with `curl` (wrapped in `retry`) into `repos/custom-post-install/`, `chmod +x`'d, then run.
- New `run-custom-postinstall-script` make target, appended as the last step of `make all` (after the kata conditional).
- `CUSTOM_POSTINSTALL_SCRIPT` added to `ci/env.defaults` and `ci/env.template` with an empty default, so `make validate-env-files` passes and the feature is fully opt-in.

## Notes

- Empty by default — a no-op when unconfigured, so existing runs are unaffected.
- Downloads land under the already-gitignored `repos/` dir; no new `.gitignore` entries.
- The existing internal `POST_INSTALL_SCRIPT` Makefile variable (`scripts/post-install.sh`) is untouched — the new name avoids any collision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ci/env.defaults
```diff
@@ -224,3 +224,14 @@ DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REPO=${DPF_HCP_PROVISIONER_OPERATOR_IMAGE_REP
 DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG=${DPF_HCP_PROVISIONER_OPERATOR_IMAGE_TAG:-latest}
 DPFHCPPROVISIONER_PULL_SECRET_NAME=${DPFHCPPROVISIONER_PULL_SECRET_NAME:-my-pull-secret}
 DPFHCPPROVISIONER_SSH_SECRET_NAME=${DPFHCPPROVISIONER_SSH_SECRET_NAME:-my-ssh-key}
+
+# Custom manifests from git (applied to the management cluster at the end of `make all`)
+# Two slots supported. Leave CUSTOM_MANIFESTS_<N> empty to disable a slot.
+# _DIR (optional) limits apply to a subdirectory; _REF (optional) pins a branch/tag/commit.
+# Repos are assumed to be pre-authenticated (host git credentials / SSH agent).
+CUSTOM_MANIFESTS_1=${CUSTOM_MANIFESTS_1:-}
```
tsorya: Can we make it simpler? Create a folder for custom manifests and apply all of them, if no manifests do nothing else apply all, will make this code couple of lines and remove all those vars

eelgaev: Folder? Right now it will point to external git repos

eelgaev:
@tsorya I simplified the PR

tsorya (approved):

tsorya:
> /cherry-pick release-4.22

wait, lets first run it on main for sometime , no need to push it ro 4.22

PR #406, opened 2026-09-18
Set TFT runtime class when Kata is enabled

tsorya:
## Summary
- Add optional `runtime_class_name` to TFT config when `KATA_ENABLED=true`
- Generate TFT config via `update_file_multi_replace` instead of direct `sed` calls
- Omit the runtime class field entirely when Kata is disabled

## Test plan
- [ ] Run `make tft-setup` and `scripts/traffic-flow-tests.sh generate-config` with `KATA_ENABLED=false`; verify generated `tft-config.yaml` has no `runtime_class_name`
- [ ] Run the same with `KATA_ENABLED=true`; verify generated config includes `runtime_class_name: "kata-coldplug"` (or configured `KATA_RUNTIME_CLASS`)
- [ ] Run traffic flow tests on a Kata-enabled cluster and confirm TFT pods use the kata runtime class


josecastillolema:
So the idea is that when kata is enabled it will run each test twice right?
One "normal" execution and then the kata-enabled one?

